PR #6, opened 2015-02-02
Add a PrometheusSink for using http://prometheus.io/

### LINKED ISSUE #4: Prometheus sink
I'm working on a sink for Prometheus (http://prometheus.github.io). Since it depends on the Prometheus client library, I'm wondering whether you'd merge it into the go-metrics package directly (which currently has no dependencies except for the std library) or whether I should put it into a separate package...?

stapelberg:
fixes #4

### DIFF
--- a/prometheus.go
+++ b/prometheus.go
@@ -1,0 +1,84 @@
+package metrics
+
+import (
+	"strings"
+	"sync"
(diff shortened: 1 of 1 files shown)


armon:
Thanks!

PR #7, opened 2015-03-14
prometheus: replace not allowed character - with _ in metric names

stapelberg:
This is necessary when using hashicorp/raft on a machine whose hostname
contains dashes, like the travis build nodes.


armon:
@stapelberg Can you move the logic into a bit more sequential instead of chaining so many of the replaces? It is getting hard to read. Thanks!

stapelberg:
Done. Let me know if you want me to squash the changes into a single commit or whether you want to leave it as-is to preserve history of this pull request.

armon:
Thanks!

PR #11, opened 2015-03-22
Only build PrometheusSink on Go >=1.3.

juliusv:
This fixes https://github.com/prometheus/client_golang/issues/102

### DIFF
--- a/prometheus.go
+++ b/prometheus.go
@@ -1,3 +1,5 @@
+// +build go1.3
+
 package metrics
 
 import (


dachary:
cute :-) :+1:

armon:
Will this build _only_ on Go1.3 or on anything >= 1.3?

armon:
I'm also tempted to move this to a sub-package to avoid breaking builds.

juliusv:
This will build on anything >=1.3. See http://golang.org/pkg/go/build/:

    - ...
    - "go1.3", from Go version 1.3 onward
    - ...

Using sub-packages would work too, but this already removes the entire Prometheus file from anyone's build when they are using Go <1.3.

armon:
Cool, merging for now. We can deal with sub-packages later.

PR #12, opened 2015-05-16
adds hostname support to other metric types

spheromak:
Only Gauge types had tests/implementation for Hostname support. This fixes that.

### DIFF
--- a/metrics.go
+++ b/metrics.go
@@ -19,6 +19,9 @@ func (m *Metrics) SetGauge(key []string, val float32) {
 }
 
 func (m *Metrics) EmitKey(key []string, val float32) {
+	if m.HostName != "" && m.EnableHostname {
+		key = insert(0, m.HostName, key)
+	}
 	if m.EnableTypePrefix {
 		key = insert(0, "kv", key)
 	}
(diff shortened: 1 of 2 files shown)


fujin:
:+1:

armon:
For most metric types you don't really want that. The sample / counter types are designed to be aggregated across a service. If you inject the hostname you prevent the rollups from happening. It makes sense for gauges where they are not aggregations, but single point measurements.

What is the use case for this change? For these metrics if you really wanted hostname it makes sense to duplicate the metric and emit it twice, once with the hostname and once without.

spheromak:
@armon my specific case was for IncrCounter. Counting some message buss things on a per-host basis. My expectation was that those metrics would be submitted with hostname when I set that to on. 

If I didn't want it i wouldn't set it :)

armon:
@spheromak The problem is the way this PR is implemented, all of our existing clients would break. Since now every counter is namespaced by host where that was not happening previously.

spheromak:
@armon Understood. Is you want to make this an option per metric type ?

armon:
@spheromak That works. We should have the defaults be backwards compatible, but allow a per-metric type setting.

jondot:
my 2cents - this should be left to the user (and would probably be removed from Gauge). If must - you can define a "prefix" and "postfix" to be given by the user. In any case, let the user specify and resolve the hostname as well as the prefix.

spheromak:
@jondot I agree. Either way it should be consistent.  Dropping the hostname prepending altogether is probably best.  Removes code, and makes every metric setup consistent.

PR #13, opened 2015-05-21
add LastUpdated field to AggregateSample

kung-foo:
I frequently find it necessary to keep track of when the last sample was added in an `IntervalMetric`. This PR adds a `LastUpdated` field to `AggregateSample` and is updated whenever `Ingest()` is called.

```go
s := im.Samples["foo.bar"]
rate := (s.Max - s.Min)/(s.LastUpdated.Sub(im.Interval).Seconds())
```

### DIFF
--- a/inmem.go
+++ b/inmem.go
@@ -65,11 +65,12 @@ func NewIntervalMetrics(intv time.Time) *IntervalMetrics {
 // AggregateSample is used to hold aggregate metrics
 // about a sample
 type AggregateSample struct {
-	Count int     // The count of emitted pairs
-	Sum   float64 // The sum of values
-	SumSq float64 // The sum of squared values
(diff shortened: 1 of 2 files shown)


armon:
@kung-foo Is the granularity of the `Interval` not enough for this? e.g. we know the metric was last updated within that `Interval` period

kung-foo:
For small intervals, yes, the base granularity is sufficient. But if I was using a 60 minute window, then knowing when the last sample is valuable (for example, cpu usage over time). Also, if the samples arrive in a haphazard manner (rather than, say, a consistent 1Hz), then this provides a simple way to track that.

armon:
@kung-foo Understood. Could you please add tests for the changes?

kung-foo:
will do

PR #19, opened 2015-07-28
Add a DogStatsd sink

talwai:
This adds a very basic sink for [DogStatsd](http://docs.datadoghq.com/guides/dogstatsd/) implementing the MetricSink
interface.
It has two main features:
1. Using the client at https://github.com/DataDog/datadog-go/statsd it
allows global tags to be assigned to all metrics sent into the sink
2. When necessary, it splices out the hostname from incoming metric
names, and optinally delegates host tagging to the Dogstatsd server. This allows
for much saner grouping by host in the Datadog backend.

### DIFF
--- a/dogstatsd.go
+++ b/dogstatsd.go
@@ -1,0 +1,115 @@
+package metrics
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/Datadog/datadog-go/statsd"
+)
+
+// HostnameGetter is a generic function to retrieve a hostname
+type HostnameGetter func() string
+
+// DogStatsdSink provides a MetricSink that can be used
+// with a dogstatsd server. It utilizes the Dogstatsd client at github.com/Datadog/datadog-go/statsd
+type DogStatsdSink struct {
+	client            *statsd.Client
+	hostnameGetter    HostnameGetter
+	propagateHostname bool
+}
+
+func getHostname() string {
+	conf, _ := GetConfig()
+	return conf.HostName
+}
+
+// NewDogStatsdSink is used to create a new DogStatsdSink with sane defaults
+func NewDogStatsdSink(addr string) (*DogStatsdSink, error) {
+	client, err := statsd.New(addr)
+	if err != nil {
+		return nil, err
+	}
+
+	sink := &DogStatsdSink{
+		client:            client,
+		hostnameGetter:    getHostname, // Defaults to reading from the global configuration `HostName`
+		propagateHostname: false,
+	}
+	return sink, nil
+}
+
+func (s *DogStatsdSink) setTags(tags []string) {
+	s.client.Tags = tags
+}
+
+func (s *DogStatsdSink) enableHostnamePropagation() {
+	// Forces a Dogstatsd `host` tag with the value specified by `globalMetrics.Config.HostName`
+	// Since the go-metrics package has its own mechanism for attaching a hostname to metrics,
+	// setting the `propagateHostname` flag ensures that `globalMetrics.Config.HostName` overrides the host tag naively set by the DogStatsd server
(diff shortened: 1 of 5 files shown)


talwai:
@armon - ping. Any thoughts on this?

armon:
@talwai Could you move the sink into a sub-package called `datadog`? That way you do not add the data dog statsd client lib as a dependency if not used.

dogstatsd.go
```diff
@@ -0,0 +1,115 @@
...
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/Datadog/datadog-go/statsd"
+)
+
+// HostnameGetter is a generic function to retrieve a hostname
+type HostnameGetter func() string
```
armon: Why is this exported?

dogstatsd.go
```diff
@@ -0,0 +1,115 @@
...
+
+// DogStatsdSink provides a MetricSink that can be used
+// with a dogstatsd server. It utilizes the Dogstatsd client at github.com/Datadog/datadog-go/statsd
+type DogStatsdSink struct {
+	client            *statsd.Client
+	hostnameGetter    HostnameGetter
```
armon: Can the hostname just be a static value? Seems odd for it to change

dogstatsd.go
```diff
@@ -0,0 +1,115 @@
...
+	propagateHostname bool
+}
+
+func getHostname() string {
+	conf, _ := GetConfig()
+	return conf.HostName
+}
+
+// NewDogStatsdSink is used to create a new DogStatsdSink with sane defaults
+func NewDogStatsdSink(addr string) (*DogStatsdSink, error) {
```
armon: Just pass in the hostname here, or use os.Hostname()

dogstatsd.go
```diff
@@ -0,0 +1,115 @@
...
+
+// DogStatsdSink provides a MetricSink that can be used
+// with a dogstatsd server. It utilizes the Dogstatsd client at github.com/Datadog/datadog-go/statsd
+type DogStatsdSink struct {
+	client            *statsd.Client
+	hostnameGetter    HostnameGetter
+	propagateHostname bool
+}
+
+func getHostname() string {
```
armon: Not needed most likely

dogstatsd.go
```diff
@@ -0,0 +1,115 @@
...
+
+	sink := &DogStatsdSink{
+		client:            client,
+		hostnameGetter:    getHostname, // Defaults to reading from the global configuration `HostName`
+		propagateHostname: false,
+	}
+	return sink, nil
+}
+
+func (s *DogStatsdSink) setTags(tags []string) {
```
armon: What is this function for?

dogstatsd.go
```diff
@@ -0,0 +1,115 @@
...
+	s.client.Tags = tags
+}
+
+func (s *DogStatsdSink) enableHostnamePropagation() {
```
armon: Similarly, this function seems weird

start.go
```diff
@@ -73,6 +74,14 @@ func NewGlobal(conf *Config, sink MetricSink) (*Metrics, error) {
 	return metrics, err
 }
 
+// GetConfig returns the Config for the globalMetrics instance if available
+func GetConfig() (*Config, error) {
```
armon: I think it should just return *Config which may be nil, no need for an error return.

talwai:
@armon Sorry for the delayed response. I think I've addressed the majority of your issues in code and added more documentation around the rest to explain my approach. Let me know how this looks. Also I moved the sink into a `datadog` subpackage as you suggested

armon:
Thanks!

PR #20, opened 2015-09-23
Prepend the hostname to every metric with `EnableHostName` set

jsternberg:
For some reason, the `EnableHostName` option exists, but isn't used
except for `SetGauge`. This causes most metrics to be in the wrong
location.

Expanding this feature to work for the rest of the functions.

### DIFF
--- a/metrics.go
+++ b/metrics.go
@@ -6,11 +6,11 @@ import (
 )
 
 func (m *Metrics) SetGauge(key []string, val float32) {
-	if m.HostName != "" && m.EnableHostname {
-		key = insert(0, m.HostName, key)
-	}
 	if m.EnableTypePrefix {
 		key = insert(0, "gauge", key)
+	}
+	if m.HostName != "" && m.EnableHostname {
+		key = insert(0, m.HostName, key)
 	}
 	if m.ServiceName != "" {
 		key = insert(0, m.ServiceName, key)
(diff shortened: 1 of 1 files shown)


jsternberg:
Specifically adding this because the metrics sent by Consul and Serf don't follow the same pattern. Some are prepended with the hostname and some aren't.

We would like to be able to organize metrics by instance. This seems to be intended from the code.

jsternberg:
I noticed the original change has it so the type prefix is before the hostname. It makes more sense to have the hostname before the type prefix since all of the metrics are organized under the hostname and it makes it easier to use in tools like Grafana.

jsternberg:
Sorry, somehow I'm still having it report metrics where it reports to statsite by repeating the service name and the type prefix, even though those are both disabled. Please don't merge yet since I don't know why it isn't working the way I think it should be yet.

jsternberg:
The issue I encountered that caused the previous commit was because of the statsite configuration (specifically I forgot to turn off `use_type_prefix` and the configuration that got templated was different than what I anticipated).

The second commit is still valid though since having the hostname before the type prefix is still really helpful for organization.

jefferai:
Chiming in, I'd really really like to be able to set the hostname option per metric. Some need to have hostnames to be meaningful, some are very hard to make use of when hostnames are included.

PR #23, opened 2015-12-03
add datadog tags

josler:
Add to DatadogStatsdSink API to provide tags.

http://docs.datadoghq.com/guides/dogstatsd/#tags

The code was calling methods with tags on them anyway, this just exposes these through an extension to the API.


armon:
Thanks!

PR #26, opened 2016-02-17
Added Throughput caclulation

vlkv:
Hi!
Can you accept this code change? I'd like to use go-metrics in my project, but I need a few little improvements... Maybe will make more pull requests.


armon:
@vlkv I would replace `Throughput` with `Rate` and also this needs tests prior to merging.

vlkv:
I'll rename Throughput to Rate (it's a better name, I agree). Please take a look at InmemFileDumper, in this branch.

I'll try to add tests in near future

vlkv:
Added general tests, please check

vlkv:
Any activity, please...

inmem.go
```diff
@@ -119,10 +123,12 @@ func (a *AggregateSample) String() string {
 // NewInmemSink is used to construct a new in-memory sink.
 // Uses an aggregation interval and maximum retention period.
 func NewInmemSink(interval, retain time.Duration) *InmemSink {
+	rateTimeUnit := time.Second
 	i := &InmemSink{
 		interval:     interval,
 		retain:       retain,
 		maxIntervals: int(retain / interval),
+		rateDenom: float64(interval.Nanoseconds()) / float64(rateTimeUnit.Nanoseconds()),
```
armon: Probably cleaner to just do `float64(interval / time.Second)` given the interval is already a `time.Duration`

inmem_file_dumper.go
```diff
@@ -0,0 +1,223 @@
...
+	Mean
+	Min
+	Max
+	Sum
+	SumSq
+	Stddev
+	LastUpdated
+)
+
+func CreateASParam(type_ AggregateSampleParamType) AggregateSampleParam {
```
armon: Please remove this entire file. I would recommend moving this to your own library if its something you want.

armon:
@vlkv I've left some feedback. I'd ask that you remove `inmem_file_dumper.go` entirely, as its not something I want to support in this package, but you should be able to cleanly move it into a separate library.

vlkv:
OK, I've removed inmem_file_dumper.go and modified formula

armon:
Thanks!

PR #29, opened 2016-04-19
Revert commit b79a89c to fix tests

tpot:
The code cleanup commit in b79a89c breaks the unit tests.  In particular rateDenom becomes zero if nanoseconds are not used for the calculation causing the Rate to become +Inf.  This may effect non-test code as well.


armon:
Thanks!

PR #30, opened 2016-04-26
Replace = by _ in Prometheus metrics

discordianfish:
Otherwise go-metrics panics which can have severe impact. Would like to get this little fix in quickly. Afterwards I'll look into how to do the PrometheusSink a bit nicer.


armon:
Thanks!

PR #31, opened 2016-05-02
Fix counter metric types

discordianfish:
The following metrics are counters, not gauges:

- stats.Mallocs
- stats.Frees
- stats.PauseTotalNs
- stats.NumGC


armon:
@discordianfish Counters have the behavior of summing, where a gauge will retain the last value, which makes more since since things like Total Pause and NumGC are already an aggregate, otherwise you would get a double summing.

discordianfish:
Ah,  you mean setCounter actually increments the counter?

armon:
@discordianfish Counters will accumulate typically, so metrics like `Sum` and `Count` will do rollups, where Gauge is a point value, which makes more sense for this type of information.

discordianfish:
Okay, I guess this is then something statsd-specific. At least in Prometheus' sense they are counters, not gauges. Confusing.. :)

PR #32, opened 2016-07-11
Integrate Circonus

maier:
Add a Circonus metrics sink.


maier:
switching to typed vars

PR #44, opened 2017-02-07
PoC exposing tagging funcs to enable DataDog tagging in nomad

kris-runzer:
I've just done some scratch work on implementing the missing DataDog tagging for Nomad.  Would a PR such as this (obviously missing tests ATM) be an acceptable solution to move further upstream into Nomad?

Related issue: https://github.com/hashicorp/nomad/issues/2110

### DIFF
--- a/metrics.go
+++ b/metrics.go
@@ -61,6 +61,39 @@ func (m *Metrics) MeasureSince(key []string, start time.Time) {
 	m.sink.AddSample(key, msec)
 }
 
+func (m *Metrics) SetGaugeWithTags(key []string, val float32, tags []string) {
+	if m.HostName != "" && m.EnableHostname {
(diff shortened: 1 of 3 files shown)


dadgar:
@kris-runzer If you want to tackle this we would actually want to make an API that is compatible between sinks that both support and do not support tagging. The thought is to take a key name and tags and if the sink doesn't support tags, the tags can be collapsed into the key name.

aantono:
Looks like this one can be closed, since this functionality is already done via https://github.com/armon/go-metrics/commit/ded85ed431a7aee3f3af79f082b704d948058f64

dadgar:
@aantono Thanks for the ping. @kris-runzer Thanks for the initial PoC!

PR #45, opened 2017-02-14
ignore feature using syscall for appengine users

tutuming:
go-metrics is imported from goa ( https://github.com/goadesign/goa ) and some users (including me) use appengine to serve APIs.

Build for appengine failed because import of `syscall` package is not allowed.
So for the appengine users, ignoring the InmemSignal feature make them happy.

### DIFF
--- a/const_unix.go
+++ b/const_unix.go
@@ -1,4 +1,5 @@
 // +build !windows
+// +build !appengine
 
 package metrics
 
--- a/const_windows.go
+++ b/const_windows.go
@@ -1,4 +1,5 @@
 // +build windows
+// +build !appengine
 
 package metrics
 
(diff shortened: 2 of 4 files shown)


tikasan:
I hope this request will be merged soon

ryanuber:
This seems fine to me. Can you also add a note to the README that the InmemSignal is not supported on this platform?

tikasan:
Go-app-builder: Failed parsing input: parser: bad import "syscall" in ...
This problem is a problem occurring in GAE which is PaaS of GCP.
Two packages, syscall and unsafe, are not allowed to be used on GAE.
So, you need a description not included in the build.

The target files are as follows.
- const_unix.go 
- const_windows.go 
- inmem_signal.go 
- inmem_signal_test.go 

We have created a request.
Verification please.
https://github.com/armon/go-metrics/pull/47

PR #46, opened 2017-03-17
Add an interface declaration for Metrics

bobbytables:
Also fixed some golint errors on the package level functions.

### Purpose

Right now we use `*metrics.Metrics` as the type when we initialize things like servers or daemons with this package. This is fine in tests if you use a blackhole sink but it would be nice to just have something that guarantees the interface that we can implement ourselves for testing purposes. Since `MockSink` isn't exported. Alternatively we could just export the MockSink (its declared in a test file).

### DIFF
--- a/start.go
+++ b/start.go
@@ -24,6 +24,16 @@ type Metrics struct {
 	sink      MetricSink
 }
 
+// Interface defines a simple metrics interface that can be implemented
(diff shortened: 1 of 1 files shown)


bobbytables:
Ping, any thoughts on this?

PR #48, opened 2017-05-26
Introduce unified URL-based sink factory function

phinze:
The `NewMetricSinkFromURL` function allows a single string input to
configure any of the supported sink backends. The various pieces of the
URL are interpreted on a per-sink basis, and the generic `MetricSink`
interface is returned.

This is convenient for downstream users of go-metrics that want to
expose sink-agnostic configuration options, but for whom a more
complicated (e.g. HCL-based) config language is too heavy.

Currently the statsd, statsite, and inmem sinks are supported. Support
for the remaining sinks can be added as a follow up.

### DIFF
--- a/inmem.go
+++ b/inmem.go
@@ -3,6 +3,7 @@ package metrics
 import (
 	"fmt"
 	"math"
+	"net/url"
 	"strings"
 	"sync"
 	"time"
@@ -25,7 +26,7 @@ type InmemSink struct {
 	// intervals is a slice of the retained intervals
 	intervals    []*IntervalMetrics
 	intervalLock sync.RWMutex
-	
+
 	rateDenom float64
 }
 
@@ -68,7 +69,7 @@ func NewIntervalMetrics(intv time.Time) *IntervalMetrics {
 // about a sample
 type AggregateSample struct {
 	Count       int       // The count of emitted pairs
-	Rate	        float64   // The count of emitted pairs per time unit (usually 1 second)
+	Rate        float64   // The count of emitted pairs per time unit (usually 1 second)
 	Sum         float64   // The sum of values
 	SumSq       float64   // The sum of squared values
 	Min         float64   // Minimum value
@@ -105,7 +106,7 @@ func (a *AggregateSample) Ingest(v float64, rateDenom float64) {
 	if v > a.Max || a.Count == 1 {
 		a.Max = v
 	}
-	a.Rate = float64(a.Count)/rateDenom
+	a.Rate = float64(a.Count) / rateDenom
 	a.LastUpdated = time.Now()
 }
 
@@ -120,6 +121,27 @@ func (a *AggregateSample) String() string {
 	}
 }
 
+// NewInmemSinkFromURL creates an InmemSink from a URL. It is used
+// (and tested) from NewMetricSinkFromURL.
+func NewInmemSinkFromURL(u *url.URL) (MetricSink, error) {
+	params := u.Query()
+	if len(params["interval"]) < 1 {
+		return nil, fmt.Errorf("missing interval query param")
+	}
+	interval, err := time.ParseDuration(params["interval"][0])
+	if err != nil {
+		return nil, err
+	}
+	if len(params["duration"]) < 1 {
+		return nil, fmt.Errorf("missing duration query param")
+	}
+	duration, err := time.ParseDuration(params["duration"][0])
+	if err != nil {
+		return nil, err
+	}
+	return NewInmemSink(interval, duration), nil
+}
+
 // NewInmemSink is used to construct a new in-memory sink.
 // Uses an aggregation interval and maximum retention period.
 func NewInmemSink(interval, retain time.Duration) *InmemSink {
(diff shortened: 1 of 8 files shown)


phinze:
ping @ryanuber / @evanphx / @mmcquillan

inmem.go
```diff
@@ -120,6 +121,27 @@ func (a *AggregateSample) String() string {
 	}
 }
 
+// NewInmemSinkFromURL creates an InmemSink from a URL. It is used
+// (and tested) from NewMetricSinkFromURL.
+func NewInmemSinkFromURL(u *url.URL) (MetricSink, error) {
+	params := u.Query()
+	if len(params["interval"]) < 1 {
```
ryanuber: You can use `.Get(key)` on a `url.Values` instead of checking length + indexing to the 0th thing. I would just do that and check against `""` to make sure it was set.

https://golang.org/pkg/net/url/#Values.Get

ryanuber: You could also collapse this with the `ParseDuration` call and just blindly try to parse whatever `Values.Get()` gives you. That way if it was empty you would just get an error like `invalid interval: ""`.

inmem.go
```diff
@@ -120,6 +121,27 @@ func (a *AggregateSample) String() string {
...
+		return nil, fmt.Errorf("missing interval query param")
+	}
+	interval, err := time.ParseDuration(params["interval"][0])
+	if err != nil {
+		return nil, err
+	}
+	if len(params["duration"]) < 1 {
```
ryanuber: Same comments as above for the `duration`

sink.go
```diff
@@ -50,3 +56,44 @@ func (fh FanoutSink) AddSample(key []string, val float32) {
...
+// "inmem://" - Initializes an InmemSink. The host and port are ignored. The
+// "interval" and "duration" query parameters must be specified with valid
+// durations, see NewInmemSink for details.
+func NewMetricSinkFromURL(urlStr string) (MetricSink, error) {
+	u, err := url.Parse(urlStr)
+	if err != nil {
+		return nil, err
+	}
+
+	sinkName := strings.ToLower(u.Scheme)
```
ryanuber: Should we accept things like `STATSD://...`? That just screams misuse to me. If there is a reason to keep it, we should also test it.

sink_test.go
```diff
@@ -118,3 +120,77 @@ func TestFanoutSink_Sample(t *testing.T) {
...
+				if !strings.Contains(err.Error(), "unrecognized sink name: \"notasink\"") {
+					return fmt.Errorf("unexpected kind of err: %s", err)
+				}
+				return nil
+			},
+		},
+	}
+
+	for name, tc := range cases {
+		output, err := NewMetricSinkFromURL(tc.Input)
```
ryanuber: This is a good candidate test to use `t.Run()` - that way even if one of these fails, they still all run so you get the full view of what all of the tests actually return.

sink_test.go
```diff
@@ -118,3 +120,77 @@ func TestFanoutSink_Sample(t *testing.T) {
 		t.Fatalf("val not equal")
 	}
 }
+
+func TestNewMetricSinkFromURL(t *testing.T) {
```
ryanuber: I think we need to restructure this test a bit. While it does cover the URL-enabled sinks broadly, we are skipping some of the depth. I would just have this function ensure that the output is a sink of the expected type, and then test the `NewXXXSinkFromURL`'s in isolation. This would also make it clearer because they all have different possibilities of returned errors.

sink_test.go
```diff
@@ -118,3 +120,77 @@ func TestFanoutSink_Sample(t *testing.T) {
...
+		},
+		"inmem": {
+			Input: "inmem://?interval=30s&duration=30s",
+			Check: func(ms MetricSink, err error) error {
+				if err != nil {
+					return fmt.Errorf("unexpected err: %s", err)
+				}
+				if _, ok := ms.(*InmemSink); !ok {
+					return fmt.Errorf("Response is not a *InmemSink: %#v", ms)
+				}
```
ryanuber: We should test invalid and missing duration/interval.

inmem.go
phinze: Ah nice! Will do

sink.go
phinze: True - will just drop it.

sink_test.go
phinze: I was starting to fan out the per-sink factory tests but they started to feel boiler-platey. I was on the fence though so with this nudge I'll drop 'em in! <3

sink_test.go
phinze: Ooo! I'll look into it.

sink_test.go
phinze: Agreed will do.

phinze:
Ok @ryanuber PTAL!

inmem_test.go
```diff
@@ -99,9 +101,78 @@ func TestInmemSink(t *testing.T) {
...
+			expectInterval: duration(t, "11s"),
+			expectRetain:   duration(t, "22s"),
+		},
+		{
+			desc:      "interval is required",
+			input:     "inmem://?retain=22s",
+			expectErr: "Bad 'interval' param",
+		},
+		{
+			desc:      "interval must be a duratgion",
```
ryanuber: typo!

Issue #49, opened 2017-05-30
Fix race condition in start.go


magiconair:
The initialization of the `globalMetrics` field is racy and this is exposed in parallel tests when run with the race detector. Wrapping the `globalMetrics` field in an `atomic.Value` does not have a performance impact.

```
$ go test -bench GlobalMetrics
Benchmark_GlobalMetrics_Direct/direct-8         	 5000000	       282 ns/op
Benchmark_GlobalMetrics_Direct/atomic.Value-8   	10000000	       352 ns/op
$ go test -bench GlobalMetrics
Benchmark_GlobalMetrics_Direct/direct-8         	 5000000	       326 ns/op
Benchmark_GlobalMetrics_Direct/atomic.Value-8   	10000000	       251 ns/op
```

slackpad:
Fixed in #50.

PR #50, opened 2017-05-30
Issue #49: Fix race condition in start.go

magiconair:
Access to the globalMetrics variable is racy and causes the race
detector in parallel tests to trigger. Wrapping the var into an
atomic.Value does not have any performance impact and it fixes this
issue.


slackpad:
LGTM

PR #53, opened 2017-07-14
Added NewStatsdSinkWithPrefix function

VoyTechnology:
`NewStatsdSinkWithPrefix` allows each metric to be sent with a prefix.
Based on https://github.com/peterbourgon/g2s/pull/13

I have noticed that `DogStatsdSink` uses normal `NewDogStatsdSink` and then `SetTags`, so I have no problem with changing this sink to use `SetPrefix` if you think the current solution is not optimal.

### DIFF
--- a/statsd.go
+++ b/statsd.go
@@ -21,6 +21,7 @@ const (
 // only UDP packets, while StatsiteSink uses TCP.
 type StatsdSink struct {
 	addr        string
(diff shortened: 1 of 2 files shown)


VoyTechnology:
Apologies. It turns out Consul was setting the prefix anyways in a weird way. Closing.

PR #54, opened 2017-08-02
Add support for labeled metrics

kyhavlov:
This adds new global functions for submitting metrics with key/value labels, as well as an endpoint function for displaying metrics from the most recent finished interval. Also modifies existing sinks to handle tagged metrics where applicable.

The endpoint's output looks like this when serialized to json:
```json
{
    "Timestamp": "2017-08-02 01:06:00 -0700 PDT",
    "Gauges": [
        {
            "Name": "consul.consul.session_ttl.active",
            "Value": 0,
            "Labels": {}
        },
        {
            "Name": "consul.runtime.alloc_bytes",
            "Value": 4228152,
            "Labels": {}
        }
    ],
    "Points": [],
    "Counters": [
        {
            "Name": "consul.consul.catalog.service.query",
            "Count": 3,
            "Sum": 3,
            "Min": 1,
            "Max": 1,
            "Mean": 1,
            "Stddev": 0,
            "Labels": {
                "service": "consul"
            }
        }
    ],
    "Samples": [
        {
            "Name": "consul.serf.queue.Query",
            "Count": 20,
            "Sum": 0,
            "Min": 0,
            "Max": 0,
            "Mean": 0,
            "Stddev": 0,
            "Labels": {}
        }
    ]
}

### DIFF
--- a/circonus/circonus_test.go
+++ b/circonus/circonus_test.go
@@ -87,7 +87,7 @@ func TestSetGauge(t *testing.T) {
 		cs.Flush()
 	}()
 
-	expect := "{\"foo`bar\":{\"_type\":\"n\",\"_value\":1}}"
+	expect := "{\"foo`bar\":{\"_type\":\"n\",\"_value\":\"1\"}}"
 	actual := <-q
 
 	if actual != expect {
--- a/circonus/circonus_test.go
+++ b/circonus/circonus_test.go
@@ -12,7 +12,7 @@ import (
 func TestNewCirconusSink(t *testing.T) {
 
 	// test with invalid config (nil)
-	expectedError := errors.New("Invalid check manager configuration (no API token AND no submission url).")
+	expectedError := errors.New("invalid check manager configuration (no API token AND no submission url)")
 	_, err := NewCirconusSink(nil)
 	if err == nil || err.Error() != expectedError.Error() {
 		t.Errorf("Expected an '%#v' error, got '%#v'", expectedError, err)
--- a/datadog/dogstatsd.go
+++ b/datadog/dogstatsd.go
@@ -97,9 +97,8 @@ func (s *DogStatsdSink) AddSample(key []string, val float32) {
 	s.AddSampleWithLabels(key, val, nil)
 }
 
-// The following ...WithTags methods correspond to Datadog's Tag extension to Statsd.
+// The following ...WithLabels methods correspond to Datadog's Tag extension to Statsd.
 // http://docs.datadoghq.com/guides/dogstatsd/#tags
-
 func (s *DogStatsdSink) SetGaugeWithLabels(key []string, val float32, labels []metrics.Label) {
 	flatKey, tags := s.getFlatkeyAndCombinedLabels(key, labels)
 	rate := 1.0
--- a/datadog/dogstatsd.go
+++ b/datadog/dogstatsd.go
@@ -69,6 +69,7 @@ func (s *DogStatsdSink) parseKey(key []string) ([]string, []metrics.Label) {
 	for i, el := range key {
 		if el == hostName {
 			key = append(key[:i], key[i+1:]...)
+			break
 		}
 	}
 
--- a/datadog/dogstatsd.go
+++ b/datadog/dogstatsd.go
@@ -1,10 +1,12 @@
 package datadog
 
 import (
-	"fmt"
 	"strings"
 
+	"fmt"
+
 	"github.com/DataDog/datadog-go/statsd"
+	"github.com/armon/go-metrics"
 )
 
 // DogStatsdSink provides a MetricSink that can be used
@@ -57,11 +59,11 @@ func (s *DogStatsdSink) flattenKey(parts []string) string {
 	}, joined)
 }
 
-func (s *DogStatsdSink) parseKey(key []string) ([]string, []string) {
+func (s *DogStatsdSink) parseKey(key []string) ([]string, []metrics.Label) {
 	// Since DogStatsd supports dimensionality via tags on metric keys, this sink's approach is to splice the hostname out of the key in favor of a `host` tag
 	// The `host` tag is either forced here, or set downstream by the DogStatsd server
 
-	var tags []string
+	var labels []metrics.Label
 	hostName := s.hostName
 
 	//Splice the hostname out of the key
@@ -72,19 +74,19 @@ func (s *DogStatsdSink) parseKey(key []string) ([]string, []string) {
 	}
 
 	if s.propagateHostname {
-		tags = append(tags, fmt.Sprintf("host:%s", hostName))
+		labels = append(labels, metrics.Label{"host", hostName})
 	}
-	return key, tags
+	return key, labels
 }
 
 // Implementation of methods in the MetricSink interface
 
 func (s *DogStatsdSink) SetGauge(key []string, val float32) {
-	s.SetGaugeWithTags(key, val, []string{})
+	s.SetGaugeWithLabels(key, val, nil)
 }
 
 func (s *DogStatsdSink) IncrCounter(key []string, val float32) {
-	s.IncrCounterWithTags(key, val, []string{})
+	s.IncrCounterWithLabels(key, val, nil)
 }
 
 // EmitKey is not implemented since DogStatsd does not provide a metric type that holds an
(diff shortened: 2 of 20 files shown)

### CODE REVIEW

File: inmem.go [hunk 1]
```diff
@@ -264,3 +276,14 @@ func (i *InmemSink) flattenKey(parts []string) string {
 	joined := strings.Join(parts, ".")
 	return strings.Replace(joined, " ", "_", -1)
 }
+
+// Flattens the key for formatting along with its tags, removes spaces
```
slackpad: s/tags/labels/

inmem_endpoint.go
```diff
@@ -0,0 +1,133 @@
...
+		return nil, fmt.Errorf("no metric intervals have been initialized yet")
+	case n == 1:
+		interval = i.intervals[0]
+	default:
+		interval = i.intervals[n-2]
+	}
+
+	summary := MetricsSummary{
+		Timestamp: interval.Interval.String(),
+		Gauges:    []GaugeValue{},
```
slackpad: Are these not left `nil`-valued so they show up as empty arrays in the JSON output?

inmem_endpoint.go
```diff
@@ -0,0 +1,133 @@
...
+	// deterministic order.
+	for name, points := range interval.Points {
+		summary.Points = append(summary.Points, PointValue{name, points})
+	}
+	sort.Slice(summary.Points, func(i, j int) bool {
+		return summary.Points[i].Name < summary.Points[j].Name
+	})
+
+	for name, value := range interval.Gauges {
+		key, labels := extractLabels(name)
```
slackpad: This feels like a lot of processing. Should we just keep extra state around with each metric?

circonus/circonus_test.go
```diff
@@ -87,7 +87,7 @@ func TestSetGauge(t *testing.T) {
 		cs.Flush()
 	}()
 
-	expect := "{\"foo`bar\":{\"_type\":\"n\",\"_value\":1}}"
+	expect := "{\"foo`bar\":{\"_type\":\"n\",\"_value\":\"1\"}}"
```
kyhavlov: Yes, this test was failing without this change

inmem_endpoint.go
```diff
@@ -0,0 +1,133 @@
...
+	Name string
+	*AggregateSample
+	Mean   float64
+	Stddev float64
+	Labels map[string]string
+}
+
+// DisplayMetrics returns a summary of the metrics from the most recent finished interval.
+func (i *InmemSink) DisplayMetrics(resp http.ResponseWriter, req *http.Request) (interface{}, error) {
+	i.intervalLock.Lock()
```
kyhavlov: yep, calling Data() is better than taking the lock here

inmem_endpoint.go
kyhavlov: that's right, it seemed better to not show `null`s in the json output

circonus/circonus_test.go
slackpad: Current HEAD passes locally for me though.

inmem_endpoint.go
kyhavlov: We could change the gauges field from `map[string]float32` to `map[string]GaugeValue` and so on for the rest of the metric types. You're just referring to having to pull the labels back out of the name here and not the sorting, right?

circonus/circonus_test.go
kyhavlov: that's weird, maybe it's a vendor discrepancy? I didn't change the code getting called by this test so I thought it was broken

inmem_endpoint.go
slackpad: Yeah - it seems weird to have to parse the labels back out for display.

circonus/circonus_test.go
slackpad: OK nm - when I update my Circonus library locally this fails as well.

inmem_endpoint.go
```diff
@@ -0,0 +1,100 @@
...
+	case n == 0:
+		return nil, fmt.Errorf("no metric intervals have been initialized yet")
+	case n == 1:
+		interval = i.intervals[0]
+	default:
+		interval = i.intervals[n-2]
+	}
+
+	summary := MetricsSummary{
+		Timestamp: interval.Interval.String(),
```
slackpad: In Consul we ended up putting all of the timestamps we display into UTC so people don't have to mess with time zones - this might be good to do:

https://github.com/hashicorp/consul/blob/v0.9.0/agent/consul/acl_replication.go#L255-L258

I'm not sure if we want to shred any of the precision to get them to display uniformly (or if that matters here).

inmem.go
```diff
@@ -154,12 +154,16 @@ func NewInmemSink(interval, retain time.Duration) *InmemSink {
...
+}
+
+func (i *InmemSink) SetGaugeWithLabels(key []string, val float32, labels []Label) {
+	k := i.flattenKeyLabels(key, labels)
 	intv := i.getInterval()
 
 	intv.Lock()
 	defer intv.Unlock()
-	intv.Gauges[k] = val
+	intv.Gauges[k] = GaugeValue{Name: i.flattenKey(key), Value: val, Labels: labels}
```
slackpad: Since there's an implicit `flattenKey()` inside `flattenKeyLabels()` and this is a pretty hot path, how about:

`k, name := i.flattenKeyLabels`

And then you can just plug name into the structure?

prometheus/prometheus.go
```diff
@@ -24,24 +26,43 @@ func NewPrometheusSink() (*PrometheusSink, error) {
...
-func (p *PrometheusSink) flattenKey(parts []string) string {
-	joined := strings.Join(parts, "_")
-	joined = strings.Replace(joined, " ", "_", -1)
-	joined = strings.Replace(joined, ".", "_", -1)
-	joined = strings.Replace(joined, "-", "_", -1)
-	joined = strings.Replace(joined, "=", "_", -1)
-	return joined
+func (p *PrometheusSink) flattenKey(parts []string, labels []metrics.Label) (string, string) {
+	key := strings.Join(parts, "_")
+	key = strings.Replace(key, " ", "_", -1)
```
slackpad: This feels like a job for a regex :-)

inmem_endpoint.go
```diff
@@ -0,0 +1,101 @@
...
+	sort.Slice(summary.Points, func(i, j int) bool {
+		return summary.Points[i].Name < summary.Points[j].Name
+	})
+
+	for hash, value := range interval.Gauges {
+		value.Hash = hash
+		summary.Gauges = append(summary.Gauges, value)
+	}
+	sort.Slice(summary.Gauges, func(i, j int) bool {
+		return summary.Gauges[i].Hash < summary.Gauges[j].Hash
```
preetapan: was curious why we sort by hashcode instead of by name here..

kyhavlov: The hash is the name + labels concatenated, so we get everything back in a deterministic order. If we just sort by name and there's a metric with multiple different label values, there will be multiple entries with the same name and they end up in a non-deterministic order

preetapan: ok..calling it hash when its not actually applying a hash function is a bit unclear. I think implementing the `Stringer` interface via a String() method on the struct that concats it the way you want, to use inside sort would be nicer. Minor detail though.

datadog/dogstatsd.go
```diff
@@ -57,11 +58,11 @@ func (s *DogStatsdSink) flattenKey(parts []string) string {
 	}, joined)
 }
 
-func (s *DogStatsdSink) parseKey(key []string) ([]string, []string) {
+func (s *DogStatsdSink) parseKey(key []string) ([]string, []metrics.Label) {
 	// Since DogStatsd supports dimensionality via tags on metric keys, this sink's approach is to splice the hostname out of the key in favor of a `host` tag
 	// The `host` tag is either forced here, or set downstream by the DogStatsd server
 
-	var tags []string
+	var labels []metrics.Label
```
dadgar: On line 71, can you add a break?

metrics.go
```diff
@@ -5,7 +5,16 @@ import (
...
+type Label struct {
+	Name  string
+	Value string
+}
+
 func (m *Metrics) SetGauge(key []string, val float32) {
+	m.SetGaugeWithLabels(key, val, nil)
+}
+
+func (m *Metrics) SetGaugeWithLabels(key []string, val float32, labels []Label) {
```
dadgar: Can host name and service name be a label?

datadog/dogstatsd_test.go
```diff
@@ -53,13 +54,14 @@ var MetricSinkTests = []struct {
 	{"AddSample", []string{"sample", "thing"}, float32(4), EmptyTags, HostnameDisabled, "sample.thing:4.000000|ms"},
 	{"IncrCounter", []string{"count", "me"}, float32(3), EmptyTags, HostnameDisabled, "count.me:3|c"},
 
-	{"SetGauge", []string{"foo", "baz"}, float32(42), []string{"my_tag:my_value"}, HostnameDisabled, "foo.baz:42.000000|g|#my_tag:my_value"},
-	{"SetGauge", []string{"foo", "bar"}, float32(42), []string{"my_tag:my_value", "other_tag:other_value"}, HostnameDisabled, "foo.bar:42.000000|g|#my_tag:my_value,other_tag:other_value"},
-	{"SetGauge", []string{"foo", "bar"}, float32(42), []string{"my_tag:my_value", "other_tag:other_value"}, HostnameEnabled, "foo.bar:42.000000|g|#my_tag:my_value,other_tag:other_value,host:test_hostname"},
+	{"SetGauge", []string{"foo", "baz"}, float32(42), []metrics.Label{{"my_tag", "my_value"}}, HostnameDisabled, "foo.baz:42.000000|g|#my_tag:my_value"},
```
dadgar: Test label without value?

inmem.go
```diff
@@ -264,3 +284,14 @@ func (i *InmemSink) flattenKey(parts []string) string {
 	joined := strings.Join(parts, ".")
 	return strings.Replace(joined, " ", "_", -1)
 }
+
+// Flattens the key for formatting along with its labels, removes spaces
+func (i *InmemSink) flattenKeyLabels(parts []string, labels []Label) (string, string) {
+	key := strings.Replace(strings.Join(parts, "."), " ", "_", -1)
```
dadgar: If performance is a concern these things could be done more efficiently with a bytes buffer and https://golang.org/pkg/strings/#Replacer

inmem_endpoint.go
```diff
@@ -0,0 +1,101 @@
...
+		return nil, fmt.Errorf("no metric intervals have been initialized yet")
+	case n == 1:
+		interval = i.intervals[0]
+	default:
+		interval = i.intervals[n-2]
+	}
+
+	summary := MetricsSummary{
+		Timestamp: interval.Interval.Round(time.Second).UTC().String(),
+		Gauges:    []GaugeValue{},
```
dadgar: Pre-allocate the sizes to avoid unnecessary grows?

inmem_endpoint.go
```diff
@@ -0,0 +1,101 @@
...
+	})
+
+	summary.Counters = formatSamples(interval.Counters)
+	summary.Samples = formatSamples(interval.Samples)
+
+	return summary, nil
+}
+
+func formatSamples(source map[string]SampledValue) []SampledValue {
+	output := []SampledValue{}
```
dadgar: Pre-allocate size

inmem_endpoint.go
kyhavlov: Then we'd have to call String() multiple times, which I wanted to avoid because this is in the hot path

metrics.go
kyhavlov: That's a good idea, I can add an option to add them as labels instead of prepending them to the key

metrics.go
```diff
@@ -76,12 +89,28 @@ func (m *Metrics) MeasureSinceWithLabels(key []string, start time.Time, labels [
...
 	now := time.Now()
 	elapsed := now.Sub(start)
 	msec := float32(elapsed.Nanoseconds()) / float32(m.TimerGranularity)
 	m.sink.AddSampleWithLabels(key, msec, labels)
 }
 
+// Returns whether the metric should be allowed based on configured prefix filters
+func (m *Metrics) allowMetric(key []string) bool {
+	if m.filter == nil || m.filter.Len() == 0 {
+		return true
```
slackpad: This should probably return `m.Config.FilterDefault` so you could block all with that `false` and an empty filter.

datadog/dogstatsd.go
slackpad: ^ Missed this one.

metrics.go
```diff
@@ -25,40 +46,98 @@ func (m *Metrics) EmitKey(key []string, val float32) {
...
 	elapsed := now.Sub(start)
 	msec := float32(elapsed.Nanoseconds()) / float32(m.TimerGranularity)
-	m.sink.AddSample(key, msec)
+	m.sink.AddSampleWithLabels(key, msec, labels)
+}
+
+// Returns whether the metric should be allowed based on configured prefix filters
+func (m *Metrics) allowMetric(key []string) bool {
+	if m.filter == nil || m.filter.Len() == 0 {
+		return true
```
slackpad: This should return `m.Config.FilterDefault` so you can filter everything out if desired.

PR #57, opened 2017-08-08
Add UpdateFilter function for updating filter at runtime

kyhavlov:
This lets us do a hot reload of the filter rules.


slackpad:
LGTM

PR #60, opened 2017-08-19
Fix broken builds against Go 1.6 and 1.7

### LINKED ISSUE #59: Builds against Go 1.6 and 1.7 are broken
It seems this happened after ded85ed431a7aee3f3af79f082b704d948058f64, which started using the Go-1.8-specific sort.Slice API.

freeekanayaka:
Avoid using sort.Slice when building against Go 1.6 or 1.7, instead
using the older sort.Sort interface. Fixes #59.

### DIFF
--- a/inmem_endpoint.go
+++ b/inmem_endpoint.go
@@ -3,7 +3,6 @@ package metrics
 import (
 	"fmt"
 	"net/http"
-	"sort"
 	"time"
 )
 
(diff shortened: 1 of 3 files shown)


freeekanayaka:
Hello,

was this PR considered for merging?

I think it's a trivial fix that makes github.com/hashicorp/raft buildable again with Go 1.6, which regressed due to this go-metrics issue (I have an app that requires to be buildable with Go 1.6 and that depends on hashicorp's raft, and hence on go-metrics).

Tahnks

JelteF:
Could this be reviewed?

tgross:
Doing a little issue/PR cleanup. Obviously we don't need this one anymore. Sorry about the delay. :blush:

PR #65, opened 2017-09-05
Add registry for external sinks

### LINKED ISSUE #63: Consider adding mechanism to register alternate Sinks with NewMetricSinkFromURL
In my case, I'd like to use the provided sub-package Dogstatsd sink but without loosing the abstraction this lirbary provides.

While I could write my own config/setup method that parses URL and does something different if protocol is `dogstatsd://` (i.e. instantiates dogstatsd sink directly instead of using `NewMetricSinkFromURL`) it seems undesirable to reinvent that wheel.

## Proposal

Allow external packages to register syncs early in execution (i.e. in `func init()`) that can hook into the central mecahnism.

That might look something like this:

```go

// Existing registry
var sinkRegistry = map[string]sinkURLFactoryFunc{
	"statsd":   NewStatsdSinkFromURL,
	"statsite": NewStatsiteSinkFromURL,
	"inmem":    NewInmemSinkFromURL,
}

// Add new mutex - might be unnecessary if we document that it's only safe to register in 
// `init` methods but I haven't found spec to say those are guaranteed not to be run in 
// parallel yet either.
var registryMu sync.RWMutex

func RegisterSyncForScheme(scheme string, factory sinkURLFactoryFunc) {
    registryMu.Lock()
    defer registryMu.Unlock()

    sinkRegistry[scheme] = factory
}

// Existing func
func NewMetricSinkFromURL(urlStr string) (MetricSink, error) {
    // ...
    registryMu.RLock()
    sinkURLFactoryFunc := sinkRegistry[u.Scheme]
    registryMu.RUnlock()
    /// ...
}
```

Then the included additional sinks or custom sinks defined in outer packages can use the same method to be instantiated at runtime.

I made an issue rather than just a straight PR with that code in just to check that there is no good reason this wasn't done already. If you agree it's worthwhile @armon I can make a PR.

banks:
Fixes #63 

Comments welcome. I added registration to existing sub-package sinks. Happy to leave that out if there is a good reason to not auto-register those. I added some basic tests for those but have not actually tested end-to-end.

### DIFF
--- a/circonus/circonus.go
+++ b/circonus/circonus.go
@@ -1,8 +1,12 @@
-// Circonus Metrics Sink
-
+// Circonus Metrics Sink provides an interface to forward metrics to Circonus.
+//
+// It also registers a factory that can be invoked by using
+// `metrics.NewMetricSinkFromURL` addressed by a URL with scheme `circonus://`.
+// The rest of the URL is ignored.
 package circonus
 
 import (
+	"net/url"
 	"strings"
 
 	"github.com/armon/go-metrics"
@@ -44,6 +48,13 @@ func NewCirconusSink(cc *Config) (*CirconusSink, error) {
 	return &CirconusSink{
 		metrics: metrics,
 	}, nil
+}
+
+func init() {
+	metrics.RegisterSinkURLFactory("circonus",
+		func(u *url.URL) (metrics.MetricSink, error) {
+			return NewCirconusSink(nil)
+		})
 }
 
 // Start submitting metrics to Circonus (flush every SubmitInterval)
--- a/datadog/dogstatsd.go
+++ b/datadog/dogstatsd.go
@@ -1,7 +1,14 @@
+// This package implements a go-metrics sink that that can be used
+// with a dogstatsd server
+//
+// It also registers a factory that can be invoked by using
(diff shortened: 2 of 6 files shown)


blalor:
This looks good!

`go test` fails because of the issue addressed in #88; I cherry-picked that PR into this branch and tests pass.

blalor:
Upon further investigation, all of the sinks need to be registered; right now that only happens if the sink's package is imported in another program.  Without that, `metrics.NewMetricSinkFromURL("dogstatsd://localhost:8125")` will return an error.

banks:
@blalor that was the intention - if we default imported all the packages to avoid that then there is not much use having them as separate "optional" packages since they would always be pulled in even if not used, and would be enabled for any aoplication that uses this even if not desired just because the end-user could modify the URL. 

Requiring the extended (or any other third-party) sink to be explicitly enabled via an anonymous import make this a backward compatible change (no new features or code magically enabled for existing applications), while allowing both the extended sinks in the repo and external ones to re-use the same setup/config via url.

banks:
I agree we should maybe document that the anonymous import is needed in the comments added here and probably in the README though...

PR #67, opened 2017-09-14
Add factory for creating DogStatsd sinks from a prexisting datadog client

bobbytables:
As of right now, the datadog metric sink does not use buffered flushes, which can cause a noticeable slowdown in emitting metrics. The DogStatsd package offers a buffered version of their client, but because the client is not exported from this sink, it's not possible to set it. This change allows you to pass in a client that you initialize.

For example, it would be nice to be able to use the [buffered client](https://godoc.org/github.com/DataDog/datadog-go/statsd#NewBuffered) like so:

```go
package foo

import (
  "github.com/DataDog/datadog-go/statsd"
  "github.com/armon/go-metrics/datadog"
)

client, err := statsd.NewBuffered("localhost:8125", 1024)
if err != nil {
  panic(err)
}

sink := datadog.NewDogStatsdSinkFromClient(client, "hostname")
```

### DIFF
--- a/datadog/dogstatsd.go
+++ b/datadog/dogstatsd.go
@@ -30,6 +30,15 @@ func NewDogStatsdSink(addr string, hostName string) (*DogStatsdSink, error) {
 	return sink, nil
 }
 
+// NewDogStatsdSinkFromClient returns a DogStatsdSink that uses a pre-existing datadog client
+func NewDogStatsdSinkFromClient(client *statsd.Client, hostName string) *DogStatsdSink {
+	return &DogStatsdSink{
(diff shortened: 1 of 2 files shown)


bobbytables:
Any thoughts on this?

aantono:
Would it be possible, as part of this PR, to also add a getter to the `DogStatsD` sync to expose the `Client`?  Since the DogStatsD client also provides ability to send events, it would be nice to be able to just instantiate a single client instance, as part of the DogStatsD sync and reuse it externally for Event emitting purposes as well.

PR #68, opened 2017-09-19
return rate in endpoint's output

sakateka:
Hi!
I want to return the counter rate in endpoint's output.
They were filtered here https://github.com/armon/go-metrics/pull/54/commits/adf3becca609f61229a757e21f394cdce9c0e9e4#diff-6ae13e3df4402693e50a11db4212316dR73


sakateka:
I think it's more correct to show the rate of increase in the value, not the counter.
A simple example is the number of requests per second, if the counter is incremented by one, then the growth rate is correct, but if the counter is increased by a certain amount of visits, then the growth rate is not correct.

sakateka:
PTAL @kyhavlov (as creator https://github.com/armon/go-metrics/pull/54) @preetapan @slackpad @dadgar

sakateka:
Someone look here please.

preetapan:
@sakateka unless I am mistaken, the rate was removed because you have all the information to calculate it correctly by looking at two consecutive pairs of <LastUpdated,Count>.  @kyhavlov ?

sakateka:
LastUpdated also filtered from json output https://github.com/armon/go-metrics/commit/adf3becca609f61229a757e21f394cdce9c0e9e4#diff-6ae13e3df4402693e50a11db4212316dR78
I not only return the `Rate`, but also adjust its calculation for the counters.
The `Rate` is always calculated, so include it in json output does not seem costly.
In any case, something needs to be returned to json :-)

slackpad:
Talked this over with @kyhavlov and this does look more correct, and as this wasn't exposed before should not break folks unless they were getting samples manually from the inmem sink, which seems unlikely.

PR #69, opened 2017-11-16
Handle ephemeral metrics

dadgar:
This PR changes the Sink to expire metrics after a configurable
duration. Before this change, ephemeral metrics would be tracked and
emitted forever. Not only does this emit undesirable metrics, it is a
memory leak.

### DIFF
--- a/prometheus/prometheus.go
+++ b/prometheus/prometheus.go
@@ -13,19 +13,90 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 )
 
-type PrometheusSink struct {
-	mu        sync.Mutex
-	gauges    map[string]prometheus.Gauge
-	summaries map[string]prometheus.Summary
-	counters  map[string]prometheus.Counter
+var (
+	// DefaultPrometheusOpts is the default set of options used when creating a
+	// PrometheusSink.
+	DefaultPrometheusOpts = PrometheusOpts{
+		Expiration: 60 * time.Second,
+	}
+)
+
+// PrometheusOpts is used to configure the Prometheus Sink
+type PrometheusOpts struct {
+	// Expiration is the duration a metric is valid for, after which it will be
+	// untracked. If the value is zero, a metric is never expired.
+	Expiration time.Duration
 }
 
+type PrometheusSink struct {
+	mu         sync.Mutex
+	gauges     map[string]prometheus.Gauge
+	summaries  map[string]prometheus.Summary
+	counters   map[string]prometheus.Counter
+	updates    map[string]time.Time
+	expiration time.Duration
+}
+
+// NewPrometheusSink creates a new PrometheusSink using the default options.
 func NewPrometheusSink() (*PrometheusSink, error) {
-	return &PrometheusSink{
-		gauges:    make(map[string]prometheus.Gauge),
-		summaries: make(map[string]prometheus.Summary),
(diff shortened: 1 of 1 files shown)


prometheus/prometheus.go
```diff
@@ -13,19 +13,90 @@ import (
...
+}
+
+// Describe is needed to meet the Collector interface.
+func (p *PrometheusSink) Describe(c chan<- *prometheus.Desc) {
+	// We must emit some description otherwise an error is returned. This
+	// description isn't shown to the user!
+	prometheus.NewGauge(prometheus.GaugeOpts{Name: "Dummy", Help: "Dummy"}).Describe(c)
+}
+
+// Collect meet the collection interface and allows us to enforce our expiration
```
slackpad: "meets"

prometheus/prometheus.go
```diff
@@ -13,19 +13,90 @@ import (
...
+// logic to clean up ephemeral metrics if their value haven't been set for a
+// duration exceeding our allowed expiration time.
+func (p *PrometheusSink) Collect(c chan<- prometheus.Metric) {
+	p.mu.Lock()
+	defer p.mu.Unlock()
+
+	now := time.Now()
+	for k, v := range p.gauges {
+		last := p.updates[k]
+		if p.expiration != 0 && !last.Add(p.expiration).After(now) {
```
slackpad: It makes more sense to me to do `last.Add(p.expiration).Before(now)` but your implementation works too.

prometheus/prometheus.go
```diff
@@ -13,19 +13,90 @@ import (
...
+			delete(p.gauges, k)
+		} else {
+			v.Collect(c)
+		}
+	}
+	for k, v := range p.counters {
+		last := p.updates[k]
+		if p.expiration != 0 && !last.Add(p.expiration).After(now) {
+			delete(p.updates, k)
+			delete(p.gauges, k)
```
slackpad: Delete from `p.counters`, not `p.gauges` here.

prometheus/prometheus.go
```diff
@@ -13,19 +13,90 @@ import (
...
+		} else {
+			v.Collect(c)
+		}
+	}
+	for k, v := range p.summaries {
+		last := p.updates[k]
+		if p.expiration != 0 && !last.Add(p.expiration).After(now) {
+			delete(p.updates, k)
+			delete(p.gauges, k)
```
slackpad: Delete from `p.summaries`, not `p.gauges` here.

prometheus/prometheus.go
slackpad: You could check for zero expiration one time and skip out vs. checking it in the loop.

prometheus/prometheus.go
```diff
@@ -13,19 +13,90 @@ import (
...
+	DefaultPrometheusOpts = PrometheusOpts{
+		Expiration: 60 * time.Second,
+	}
+)
+
+// PrometheusOpts is used to configure the Prometheus Sink
+type PrometheusOpts struct {
+	// Expiration is the duration a metric is valid for, after which it will be
+	// untracked. If the value is zero, a metric is never expired.
+	Expiration time.Duration
```
slackpad: I'd probably make this opt-in, but I can see why you'd do this to make it avoid memory leaks.

PR #70, opened 2017-11-23
Fix race inmem

sakateka:
`InmemSink.Data` return pointer to current interval (alongside with others)
`InmemSink.getExistingInterval` also return current interval.
Client can write to it while read 'current' interval from intervals returned by `InmemSink.Data()`
We need create copy for current interval.


sakateka:
@slackpad  PTAL

sakateka:
I'm testing it in a production environment, the race fixed by this PR

sakateka:
@dadgar PTAL

dadgar:
Thanks @sakateka

PR #79, opened 2018-06-21
Add Prometheus registerer configuration

povilasv:
We use a library which uses `go-metrics` and would like to have an ability to pass Prometheus registerer as an option

Example:
```
	metrics := prometheus.NewRegistry()
	metrics.MustRegister(
		version.NewCollector("thanos"),
		prometheus.NewGoCollector(),
	)

	// Memberlist uses go-metrics
	sink, err := gprom.NewPrometheusSink()
	if err != nil {
	}
	_, err = gmetrics.NewGlobal(gmetrics.DefaultConfig(cmd), sink)
	if err != nil {
            ....
	}
```


povilasv:
Hey @armon maybe you could take a look at this minor change? :)

luisdavim (approved):
LGTM

luisdavim:
I was about to open a similar PR...
@povilasv can you rebase?
@armon any reason not to merge this? Hot having this is actually blocking my use case...

armon:
Merged!

prometheus/prometheus_test.go
```diff
@@ -16,12 +17,43 @@ import (
...
+	sink, err := NewPrometheusSink()
+	if err != nil {
+		t.Fatalf("err = %v, want nil", err)
+	}
+
+	//check if register has a sink by unregistering it.
+	ok := prometheus.Unregister(sink)
+	if !ok {
+		t.Fatalf("Unregister(sink) = false, want true")
+)
```
dnephin, line 55: It looks like this is not valid Go syntax. I tried a checkout of the main branch, but this is failing to build.

I've opened a PR #114 to fix this problem.

PR #83, opened 2018-09-07
Apply the same logic to all metric type for Hostname prefixing

uepoch:
Use EnableHostname boolean in all types of metrics instead of just counter

### DIFF
--- a/start.go
+++ b/start.go
@@ -1,6 +1,7 @@
 package metrics
 
 import (
+	"fmt"
 	"os"
 	"sync"
 	"sync/atomic"
@@ -69,6 +70,11 @@ func DefaultConfig(serviceName string) *Config {
 
 // New is used to create a new instance of Metrics
 func New(conf *Config, sink MetricSink) (*Metrics, error) {
+
+	if conf.EnableHostname && conf.EnableHostnameLabel {
+		return nil, fmt.Errorf("both EnableHostname and EnableHostnameLabel can't be true at the same time")
+	}
+
 	met := &Metrics{}
 	met.Config = *conf
 	met.sink = sink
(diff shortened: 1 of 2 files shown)


uepoch:
Do you need any more information ?

start.go
```diff
@@ -69,6 +70,11 @@ func DefaultConfig(serviceName string) *Config {
 
 // New is used to create a new instance of Metrics
 func New(conf *Config, sink MetricSink) (*Metrics, error) {
+
+	if conf.EnableHostname && conf.EnableHostnameLabel {
+		return nil, fmt.Errorf("both EnableHostname and EnableHostnameLabel can't be true at the same time")
+	}
+
```
banks: Does this risk being a breaking change for clients? Is it important the are mutually exclusive? I can't think of a very food reason to use both but If you are migrating between statsd and prometheus for example you might want both being reported in the interim?

uepoch: https://github.com/armon/go-metrics/blob/master/statsd.go#L9 (not sure I linked properly)

Statsd and other sink handle labels on their own

In specific statsd case: append them in the name as a suffix, so if both were provided you could have
"localhost-something-localhost-label2

It could break clients using such configs initializers indeed but it will break the client interface anyway
This fix indeed fix the name of all the metrics except gauges if enableHostname is activated


I don't mind removing it tho: it was more of a safeguard against weird behaviors, what do you think ?

uepoch: Since the metrics initialize is "mostly" done at start-up time for app, issues will be easier to detect for existing clients that does use both, compared to not seeing the same metrics name after a `go get`
but again, if you prefer, I really don't feel it's important to keep this if

uepoch:
Hello @banks, sorry for HLing again :( , did you had the time to look at this review ?

Should I remove the safeguard ?

PR #86, opened 2018-09-24
fix some flaky tests

uepoch:
Reduced the threshold for gc_pause failure
used another comparison for inmem initial test
changed a bit formatting flag in some printf

### DIFF
--- a/inmem_test.go
+++ b/inmem_test.go
@@ -29,7 +29,7 @@ func TestInmemSink(t *testing.T) {
 	inm.AddSampleWithLabels([]string{"foo", "bar"}, 23, []Label{{"a", "b"}})
 
 	data = inm.Data()
-	if len(data) != 1 {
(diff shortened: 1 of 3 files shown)


freddygv:
This looks good to me @uepoch, I ran into similar multiple-interval flakyness with inmemsink elsewhere

uepoch:
Do you know who I should contact for this to get reviewed/merged ?

blalor:
@ryanuber @kyhavlov would you please take a look at this PR?  I haven't seen the inmem failure before, but I frequently get the gc_pause one:

```
--- FAIL: TestMetrics_EmitRuntimeStats (0.00s)
    metrics_test.go:290: bad val: [3 190024 7.1694584e+07 2249 1571 678 84427 1 84427]
```
It feels like testing gc time is going to be fragile no matter what, but lowering the threshold should make it less likely to fail.

PR #89, opened 2019-03-06
Packaging Refactoring

hugoluchessi:
# Why this refactor
- Consistency
- Leaking abstractions

## Consistency
the way to implement a new provider is not clear, because half of the providers are on the `metrics` package and half are segregated on packages.

Normalizing the way providers are created will ease this process.

## Leaking abstractions
Putting in main package the main model (Sink, Metric, Label, etc...) together with specific models (like inmemory, statsd, etc) will lead to use cross package references without being noticed.

For instance, Prometheus and the other providers are not included in  the URL Factory method, because the way there were implemented (with packages) they can't just include a method on `metrics` package referencing the specific package, this will raise a cyclic package reference.

# What was actually done
## Packaging ALL providers
All providers were moved to a folder (`providers`) in which all of them are within its own package (including `inmem`, `statsd` and `statsite`)

## Correcting some tests
tests were not running nor the code was building because of some tests.

Some other PR's were also addressing this matter like (https://github.com/armon/go-metrics/pull/86) but none were merged, I included it here.

Other tests, like of the `circonus` providers were not passing, I also corrected them.

## Segregated Factory (FromUrl)
All those factory functions in which accessed `metrics` and specific packages were segregated in a 
`factory` package in which its only responsibility is to parse the URL to the specific factory function of the provider.

This removes the parsing responsibility from specific Sink providers.

## Added Circle CI build check
Just what it says

## Renaming for better convention
- Change Sink interface name to Sinker following [Effective Go](https://golang.org/doc/effective_go.html#interface-names) good practices

- Change composite names of sink structures to only sink (InmemSink -> Sink)

## Added Prometheus unit tests
Added unit tests for prometheus for all 6 implemented methods

# Breaking changes
- Changing specific factory methods from `NewProviderSink` to `pkg.NewSink`
- Changing url factory method from `metrics` package to `factory` pkg

# Intermitent test
- Although https://github.com/armon/go-metrics/pull/86 tried to address the memstats randomization problem, it just reduced the problem, there are some false negatives.

- Statsite "server" was taking more time to listen TCP port than the code to send message and was failing to connect intermitently, I just added a 1 mili sleep to workaround this problem

But all changes are just renaming and package usage, functionality was not changed.


sexyhamster:
well, I see in your description new sinks: FanoutSink, BlackholeSink, but there are none in your code

hugoluchessi:
A little too late to answer hahaha

but it was here https://github.com/armon/go-metrics/pull/89/files#diff-346f965b89515534332cb33346f6a6a4R23

To solve this issue in the time I created this PR, I have forked this repo and this is the result https://github.com/hugoluchessi/go-metrics

I have further optimized the code and took some observability good practices like transactions and logging integrations here https://github.com/hugoluchessi/gobservable

It is not possible to reuse this PR, but sure it is valid to try redoing it in the current source.

PR #91, opened 2019-04-19
InmemSink.DisplayMetrics: use safe intervals from Data() to avoid race condition

### LINKED ISSUE #90: Race condition in in `DisplayMetrics`
We ran into a race condition in Nomad. We were calling `InmemSink.DisplayMetrics` in a tight loop in a test, and another routine called `SetGaugeWithLabels`. The result was a `fatal error: concurrent map iteration and map write`. The stack is attached. This was trivially reproduced.

The issue seems to be that `InmemSink.DisplayMetrics` retrieves the intervals from `Data()`, which is a deep copy to avoid race conditions. However, it doesn't use those: 
https://github.com/armon/go-metrics/blob/f0300d1749da6fa982027e449ec0c7a145510c3c/inmem_endpoint.go#L46-L59
Instead, it uses `i.intervals`, without a mutex.

I'm submitting a PR that uses the intervals returned from `Data()`, which was presumably the intent.

cgbaker:
resolves #90

### DIFF
--- a/inmem_endpoint.go
+++ b/inmem_endpoint.go
@@ -52,10 +52,10 @@ func (i *InmemSink) DisplayMetrics(resp http.ResponseWriter, req *http.Request) 
 		return nil, fmt.Errorf("no metric intervals have been initialized yet")
 	case n == 1:
 		// Show the current interval if it's all we have
-		interval = i.intervals[0]
+		interval = data[0]
 	default:
 		// Show the most recent finished interval if we have one
-		interval = i.intervals[n-2]
+		interval = data[n-2]
 	}
 
 	summary := MetricsSummary{


inmem_endpoint.go
```diff
@@ -52,10 +52,10 @@ func (i *InmemSink) DisplayMetrics(resp http.ResponseWriter, req *http.Request)
 		return nil, fmt.Errorf("no metric intervals have been initialized yet")
 	case n == 1:
 		// Show the current interval if it's all we have
-		interval = i.intervals[0]
+		interval = data[0]
 	default:
 		// Show the most recent finished interval if we have one
-		interval = i.intervals[n-2]
+		interval = data[n-2]
 	}
```
notnoop: I believe you need to lock the interval while reading from it too, e.g.

```suggestion
	interval.RLock()
	defer interval.RUnlock()

```

cgbaker: In the current version of the code, this is a deep copy. See `Data()`

cgbaker:
There is no test, but I'll add one.

inmem_endpoint.go
notnoop: `Data()` only deep copies the last element `n-1`in https://github.com/armon/go-metrics/blob/master/inmem.go#L241-L264 - but we may be reading element at  `n-2`.  I can think of possible scenarios where `SetGaugesWithLabels` for example gets `n-1` to manipulate it while another goroutine appends to interval array, resulting into a race of modification of `SetGaugesWithLabels` and read in this function.

Also, I see that `dumpStats()` still locks elements when iterating through elements returned from `Data()` in https://github.com/armon/go-metrics/blob/master/inmem_signal.go#L77-L100 .

Unless we have good coverage of tests that pass with `-race`, I'd suggest being defensive here.

cgbaker:
Good catch, thanks.

PR #96, opened 2019-04-29
test and fix for another race condition in DisplayMetrics

cgbaker:
there was another race condition in DisplayMetrics. `AggregateSample` wasn't deep-copied on `SampledValue` assignment, so that a lock on the parent `SampledValue` didn't protect the child `AggregateSample`. this meant that `Ingest` could write while `DisplayMetrics` was being called. 

I added to tests that document this with `-race`, then modified the copy code in `Data()` to use the new deep copy.

There is another issue in `Metrics`, where the array backing the passed `labels` slice was potentially being modified. In addition to potentially surprising users of this method, it also presents an opportunity for race conditions when a `SetGaugeWithLabels` calls rewrites the backing array while a `DisplayMetrics` call is reading it. There are two tests: the first was to verify and protect the race condition; the second was to verify that the array backing the input slice isn't mutated.

### DIFF
--- a/inmem_endpoint.go
+++ b/inmem_endpoint.go
@@ -41,6 +41,16 @@ type SampledValue struct {
 	DisplayLabels map[string]string `json:"Labels"`
 }
 
+// deepCopy allocates a new instance of AggregateSample
+func (source SampledValue) deepCopy() SampledValue {
+	dest := source
+	if source.AggregateSample != nil {
+		dest.AggregateSample = &AggregateSample{}
+		*dest.AggregateSample = *source.AggregateSample
+	}
+	return dest
+}
+
 // DisplayMetrics returns a summary of the metrics from the most recent finished interval.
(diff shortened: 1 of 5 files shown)


cgbaker:
requesting review from @notnoop

inmem_endpoint.go
```diff
@@ -41,6 +41,16 @@ type SampledValue struct {
 	DisplayLabels map[string]string `json:"Labels"`
 }
 
+// deepCopy allocates a new instance of AggregateSample
+func (source SampledValue) deepCopy() SampledValue {
+	dest := source
```
notnoop: This actually copies the struct at least two times times, e.g. one in the method call since receiver isn't a reference, and once for the assignment.  I would suggest doing something like:

```go
func (s *SampledValue) deepCopy() SampledValue {
	dest := *s
```

cgbaker: thanks, that's better!

PR #98, opened 2019-06-18
Add Set and Histogram

beekus:
This PR adds the `Set` and `Histogram` metrics for the datadog sink.

Similar to `Emit`, I added these metrics to the `MetricSink` interface and made them noop for other sinks as per [this slack thread](https://hashicorp.slack.com/archives/C2JMZ2HHD/p1560429016143300).


Eclion:
Hello!
Is there any update regarding this PR, aside from the need to rebase it?

balena-zh:
This PR is useful, but the `prometheus` package is missing implementation.

PR #99, opened 2019-08-25
Add prometheus pushserver support

stremovsky:
Hello

In our environment we have stringent security needs.

Prometheus can not pull events from the services.

We allow only opposite connection from out backend server to 3rd party prometheus server.

We implemented this patch to push events to Prometheus push server.

Thanks and best regards,
Yuli


stremovsky:
I am working on test and will submit it here.

stremovsky:
Let me know what you think now

jefferai:
Thanks!

PR #100, opened 2019-11-05
Update datadog dep

jefferai:


dnephin:
It looks like this change caused the datadog tests to start failing:

```
  dogstatsd_test.go:148: Line sample.thing:4|g|#tagkey:tagvalue does not match expected: sample.thing:4.000000|g|#tagkey:tagvalue

```

PR #102, opened 2019-11-19
Fix insert() to not modify input slice

### LINKED ISSUE #101: Metric emission methods in metrics.go modify input key slice
All the metric emission methods in `Metrics` from metrics.go are checking for the state of certain configuration flags like `EnableHostname`, and if set to true, are modifying the source `key` slice passed as an argument.

See [here](https://github.com/armon/go-metrics/blob/13f9d8e9b41c66738153b7582f1f62add8e407c8/metrics.go#L23) for one such example.

Modification of the input `key` slice causes it to be unusable when emitting it to multiple `Metrics` objects if any configuration flags which cause modification of the input `key` slice are enabled. If different metrics sinks need different configuration, it requires using multiple `Metrics` objects since the config is per-`Metrics` object.

This results in the client being responsible for creating copies of the same `key` slice for each time it emits the metric to different `Metrics` objects, which requires allocating a new `[]string`. This drastically hurts the performance of emitting metrics.

rturner3:
insert() in metrics.go is currently modifying the input slice.
Since this method is only ever used to prepend a value to the beginning of a slice,
changing the implementation to a new method insertAtBeginning() that does not modify
the input slice.

Signed-off-by: Ryan Turner <turner@uber.com>

Closes #101

### DIFF
--- a/metrics.go
+++ b/metrics.go
@@ -22,17 +22,17 @@ func (m *Metrics) SetGaugeWithLabels(key []string, val float32, labels []Label) 
 		if m.EnableHostnameLabel {
 			labels = append(labels, Label{"host", m.HostName})
 		} else if m.EnableHostname {
-			key = insert(0, m.HostName, key)
+			key = insertAtBeginning(m.HostName, key)
 		}
 	}
 	if m.EnableTypePrefix {
-		key = insert(0, "gauge", key)
+		key = insertAtBeginning("gauge", key)
 	}
 	if m.ServiceName != "" {
 		if m.EnableServiceLabel {
 			labels = append(labels, Label{"service", m.ServiceName})
 		} else {
-			key = insert(0, m.ServiceName, key)
+			key = insertAtBeginning(m.ServiceName, key)
 		}
 	}
 	allowed, labelsFiltered := m.allowMetric(key, labels)
@@ -44,10 +44,10 @@ func (m *Metrics) SetGaugeWithLabels(key []string, val float32, labels []Label) 
 
 func (m *Metrics) EmitKey(key []string, val float32) {
 	if m.EnableTypePrefix {
-		key = insert(0, "kv", key)
+		key = insertAtBeginning("kv", key)
 	}
 	if m.ServiceName != "" {
-		key = insert(0, m.ServiceName, key)
+		key = insertAtBeginning(m.ServiceName, key)
 	}
 	allowed, _ := m.allowMetric(key, nil)
 	if !allowed {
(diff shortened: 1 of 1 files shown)


amoore877:
This is also more performant, reducing the number of operations in altering the slice

rturner3:
@armon @jefferai It looks like there aren't any default reviewers for this project. I'd appreciate a review when you get a chance!

metrics.go
```diff
@@ -269,10 +269,9 @@ func (m *Metrics) emitRuntimeStats() {
...
-	s = append(s, "")
-	copy(s[i+1:], s[i:])
-	s[i] = v
-	return s
+// Creates a new slice with the provided string value as the first element
+// and the provided slice values as the remaining values.
+// Ordering of the values in the provided input slice is kept in tact in the output slice.
+func insertAtBeginning(v string, s []string) []string {
+	return append([]string{v}, s...)
 }
```
zmt: why not just leave it as a flexible insert, but correct the behavior:
```lang=go
func(i int, v string, s []string) []string {
	return append(s[:i], append([]string{v}, s[i:]...)...)
}

See https://github.com/golang/go/wiki/SliceTricks

amoore877: the thing is the flexible capability is not used; this is only called with index 0, and seems to have been as such for several years.
Unless there is a clear use case for the flexible version, we should keep it as simple as possible both for easier maintenance and for the slightly better performance (which is impactful for something like a metrics library)

@armon , was there a future use case for having the flexible version for any index?

if I'm reading this correctly, it is potentially insightful to note from @zmt's link that the current implementation is listed as a means to *avoid* creation of an additional slice when inserting at an arbitrary index, though neglects to mention the problem caused by that (the issue this PR fixes)

amoore877:
@armon @jefferai any chance of a review before new year?

jefferai:
I think @zmt is correct, since we don't know whether there was a reason for this in the future, it seems like a better option to simply correct the behavior in a small diff than change it everywhere.

> if I'm reading this correctly, it is potentially insightful to note from @zmt's link that the current implementation is listed as a means to avoid creation of an additional slice when inserting at an arbitrary index, though neglects to mention the problem caused by that

I don't really follow this statement -- in both your updated code and the code suggested by @zmt you're creating a new slice. It's just avoiding the explicit copy.

jefferai:
So I'm a bit confused here. The previous code had one copy of the elements, unless it had to allocate a new slice, in which case it would have two copies. The new code always has two copies of the elements. I'm not seeing how this is easier or more performant.

From the original description of this you're trying to change the code to not modify the underlying input slice. But since there's no insight into the capacity of a slice you can't ever be sure whether or not you're modifying it.

Wouldn't a better solution be to:
1) Explicitly make a new slice with capacity of the existing slice + 1
2) Copy all values up to i into the new slice
3) Insert the new value at i
4) Copy the remaining values into the new slice

This would have just a single copy operation for each element.

rturner3:
> So I'm a bit confused here. The previous code had one copy of the elements, unless it had to allocate a new slice, in which case it would have two copies. The new code always has two copies of the elements. I'm not seeing how this is easier or more performant.
> 
> From the original description of this you're trying to change the code to not modify the underlying input slice. But since there's no insight into the capacity of a slice you can't ever be sure whether or not you're modifying it.
> 
> Wouldn't a better solution be to:
> 
> 1. Explicitly make a new slice with capacity of the existing slice + 1
> 2. Copy all values up to i into the new slice
> 3. Insert the new value at i
> 4. Copy the remaining values into the new slice
> 
> This would have just a single copy operation for each element.

Sounds reasonable. I think I agree, what you've proposed seems like a more readable solution, and guaranteed to be the most performant.

Issue #103, opened 2019-11-20
in-mem flattenkey is non-performant


amoore877:
https://github.com/armon/go-metrics/blob/master/inmem.go#L312-L348

The current `flattenKey` logic loops through `key` elements with a `replacer`, adding the element and then a `.`. Each call is performing a replace op.

It would be more efficient to instead perform a `string.Join` on the elements, and write once only to the `replacer` (single replace op). Similar logic could be applied to `flattenKeyLabels`

mitar:
This can be closed, it has been merged?

PR #104, opened 2019-11-22
clean up and enhance perf of inmem key flatten

amoore877:
Addresses #103 

- Have in-mem `flattenKey` ops first `join` the slice elements together, *then* do a single `replace`, rather than a `replace` on every op of homebrew version of `join`
- Make the space replacer a `var` for `InMem`, rather than instantiating for every `flattenKey` call

Signed-off-by: Andrew Moore <andrew.s.moore@uber.com>

### DIFF
--- a/inmem.go
+++ b/inmem.go
@@ -10,6 +10,8 @@ import (
 	"time"
 )
 
+var spaceReplacer = strings.NewReplacer(" ", "_")
+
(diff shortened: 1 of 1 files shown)


amoore877:
@armon @jefferai @kyhavlov, would appreciate a review

PR #105, opened 2019-12-30
Logger interface

tehut:
Didn't expect this to be so large, I expect that a lot of the length is in the questions section on logger.go.

I tried including questions inline, if we hate it that's fine. Especially since this is on an open repo.  I'll definitely delete that and any comments before asking for a merge.

Also, I didn't rebase my commits down to a single commit but can do that or squash on merge, I'm fine with either.


logger/logger.go
```diff
@@ -0,0 +1,39 @@
...
+// OurLogger implements an empty logger struct to wrap the go-hclog logger
+type OurLogger struct {
+}
+
+//Logger interface is used to make stuff happen with things in places
+type Logger interface {
+	Error(string, ...interface{})
+}
+
+// Error wraps the go-hclog Error function and logs a separate error for each argument in the error object passed
```
tehut: I want to try keeping a "discarded attempts" section in my code reviews for now.  I'll flag them all to be removed before merge but I thought we could try it. I'm hoping it helps me be able to keep track of multiple attempts.  Again, may well discard.

mitar:
I do not like this because it makes it very hard to use some other logger. Maybe it would be better to use new golang's structured logger? Is structured logging used at all? Otherwise standard logger is probably good enough?

PR #107, opened 2020-02-21
Use sync.Map instead of a mutex for prometheus sink's metric maps

ncabatoff:
When under heavy load, pprof showed that the bulk of contention in Vault with a Prometheus metrics sink was on the sink's mutex.  This makes sense, as all metric updates have to go through the same lock.  When there's a lot of metric traffic typically it's the same metrics that are getting updated over and over again, meaning the typical case is that the maps are going to be read from rather than written to.  This makes them a good candidate for sync.Map.

### DIFF
--- a/prometheus/prometheus.go
+++ b/prometheus/prometheus.go
@@ -1,4 +1,4 @@
-// +build go1.3
+// +build go1.9
 
 package prometheus
 
@@ -32,11 +32,10 @@ type PrometheusOpts struct {
 }
 
 type PrometheusSink struct {
-	mu         sync.Mutex
-	gauges     map[string]prometheus.Gauge
-	summaries  map[string]prometheus.Summary
-	counters   map[string]prometheus.Counter
-	updates    map[string]time.Time
+	gauges     sync.Map
+	summaries  sync.Map
+	counters   sync.Map
+	updates    sync.Map
 	expiration time.Duration
 }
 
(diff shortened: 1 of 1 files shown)


jefferai:
This looks good, one question: will this value ever be copied anywhere? I believe it's not safe to copy a sync.Map, so if that might happen we may want to use pointers to maps.

ncabatoff:
> This looks good, one question: will this value ever be copied anywhere? I believe it's not safe to copy a sync.Map, so if that might happen we may want to use pointers to maps.

I thought about that, but it doesn't seem like it, since we only return a pointer to PrometheusSink in the constructor, and there's no code I could see that was copying those sinks by value.  I can switch to pointers to protect against someone doing that if you like?

prometheus/prometheus.go
```diff
@@ -32,11 +32,10 @@ type PrometheusOpts struct {
...
 type PrometheusSink struct {
-	mu         sync.Mutex
-	gauges     map[string]prometheus.Gauge
-	summaries  map[string]prometheus.Summary
-	counters   map[string]prometheus.Counter
-	updates    map[string]time.Time
+	gauges     sync.Map
+	summaries  sync.Map
+	counters   sync.Map
+	updates    sync.Map
```
jefferai:
```suggestion
	// If these will ever be copied, they should be converted to *sync.Map values and initialized appropriately
	gauges     sync.Map
	summaries  sync.Map
	counters   sync.Map
	updates    sync.Map
```

jefferai:
I added a suggestion to put a comment there. LGTM otherwise.

ncabatoff:
Thanks Jeff!

jefferai:
Thank you! Merged and will tag in a moment.

PR #112, opened 2020-08-30
Augment keys of all metric types in the same way

stefanbirkner:
It is unexpected that the common key prefix for some metrics types is
different. This makes it unnecessary difficult to find the metrics in
the database. This commit unifies the prefix for all metrics.

The prefix of each key is now `service name, type prefix, hostname`. All
of them are optional. `service name` is added if it is available and not
set as a label because the configuration flag `EnableServiceLabel` is
set. `type prefix` is added if it is enabled with the configuration flag
`EnableTypePrefix`. `hostname` is added if it is available and not set
as a label because `EnableHostnameLabel` is set.

The behaviour of `EmitKeys`, `IncrCounter`, `IncrCounterWithLabels`,
`AddSample`, `AddSampleWithLabels`, `MeasureSince` and
`MeasureSinceWithLabels` is changes with this commit. For `EmitKey` the
hostname will be added (details see above) but the service name will no
longer be added when `EnableServiceLabel` is set to `true`. For the
other types the hostname will be added if it is available and
`EnableHostnameLabel` is set to `false`.


stefanbirkner:
This PR contains the changes of #20 and #76.

PR #114, opened 2020-08-31
Fix syntax error and duplicate import

dnephin:
So that the tests will run. It looks like this was merged in https://github.com/armon/go-metrics/pull/79/files#r480251321


armon:
Thanks!

PR #115, opened 2020-08-31
Fix datadog tests

dnephin:
Fixes the datadog tests for the new expected value. The upgrade of github.com/DataDog/datadog-go in #100 caused the value of gauge metrics to change.

### DIFF
--- a/datadog/dogstatsd_test.go
+++ b/datadog/dogstatsd_test.go
@@ -49,15 +49,15 @@ var MetricSinkTests = []struct {
 	PropagateHostname bool
 	Expected          string
 }{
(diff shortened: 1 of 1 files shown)


dnephin:
@jefferai I believe the expected values changed with your PR #100. Does this look right to you? You can see the tests passing in #116

dnephin:
cc @banks @armon 

The `go-metrics` tests fail on master. This PR should fix the failures. Does this look right to you? 

#116 adds CI config to prevent this from happening in the future.

banks (approved):
I honestly don't know why the precision of the output changed but I don't think it matters for statsd so better that the tests pass!

Thanks!

banks:
Oh I see, it was the DD import upgrade.

PR #117, opened 2020-08-31
Remove unused interface methods

dnephin:
With the addition of labels in #54 only the new `WithLabels` methods on `Sink` are used by `go-metrics`.

My understanding is that this should be a backwards compatible change, because `Sink` is the interface accepted by `metrics.New`. Implementations of a `Sink` outside of this package would continue to work, and new one would no longer need to implement the 3 unused methods. Nothing outside of this package should call the `Sink` methods directly, they would call the methods on `Metrics`, which have not changed.

### DIFF
--- a/circonus/circonus.go
+++ b/circonus/circonus.go
@@ -56,12 +56,6 @@ func (s *CirconusSink) Flush() {
 	s.metrics.Flush()
 }
 
-// SetGauge sets value for a gauge metric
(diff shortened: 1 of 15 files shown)


tgross:
Doing a little issue/PR cleanup. This PR is not backwards compatible, because we can't know if a caller is calling ex. `IncCounter` directly on a concrete sink that it's instantiated. This would require a v2 of the library, which isn't going to happen anytime soon

PR #116, opened 2020-08-31
Add CI

dnephin:
Builds on #114 and #115

This PR adds a CircleCI config for running tests against the last 3 major versions of Go.

You can see a run against my branch here: https://app.circleci.com/pipelines/github/dnephin/go-metrics

The only thing that is missing is to add the project to CircleCI here: https://app.circleci.com/projects/

PR #118, opened 2020-09-14
Expose the global default metrics

dnephin:
This should make it easier to test code that uses go-metrics. In tests the code can be passed a different instance of Metrics, and in production code paths it can use the global Metrics returned from Default().

Without this change patching the global can cause flaky tests when other tests attempt to emit metrics to the global metrics instance.

An example of this can be see in https://github.com/hashicorp/consul/pull/8679 which uses this change to fix a flaky test.

Prior art: https://pkg.go.dev/github.com/hashicorp/go-hclog?tab=doc#Default

cc @armon @banks

### DIFF
--- a/start.go
+++ b/start.go
@@ -6,7 +6,7 @@ import (
 	"sync/atomic"
 	"time"
 
-	"github.com/hashicorp/go-immutable-radix"
+	iradix "github.com/hashicorp/go-immutable-radix"
 )
 
 // Config is used to configure metrics settings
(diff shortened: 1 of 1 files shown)


dnephin:
CI ran on my branch, but did not appear to run on this PR: https://app.circleci.com/pipelines/github/dnephin/go-metrics/6/workflows/63f98684-a0aa-4510-ad00-8739c361d07d

I think the "run on fork PRs" config needs to be enabled, along with third-party orbs (https://github.com/armon/go-metrics/pull/116#issuecomment-690354747).

PR #120, opened 2020-10-28
Add {Gauge,Summary,Counter}Definitions to PrometheusOpts to define non-expiring metrics

mkcp:
This PR introduces structs for Gauge, Summary, and Counter Definitions that can be provided to the `PrometheusSink` via its `PrometheusOpts`. The `PrometheusSink` will uses these definition slices to declare metrics when the process starts, and marks them internally to _not_ be deleted when their last-updated time passes the configured expiry. Instead, if `canDelete` is `false`, the PrometheusSink expires the metric by setting its value to NaN. When metrics aren't defined, they should behave as they always have before this PR. 

Allowing go-metrics users to define static metrics in prometheus lets us resolve a [long-requested](https://github.com/hashicorp/consul/issues/8590) [pain](https://github.com/hashicorp/consul/issues/8181) [point](https://github.com/hashicorp/consul/issues/5140) with how Consul expires metrics it has not observed within its prometheus_retention_time, which we set PrometheusSink's expiry value with. Workarounds for this have led to users configuring retention times on the order of days, weeks, months, or longer, leading to a pathological loss of precision in metrics which do not update often. Not all of these metrics remain valid even if we haven't updated them. With this change we can recommend much shorter retention times which will lead to more accurate measurements and fewer bogus stats.

As a secondary benefit, we can provide prometheus users better UX by attaching help string documentation to the metrics, as prometheus users expect.

This is a successor PR to the approach in #119 and allows ephemeral metrics in HashiCorp products to continue to exist where short-lived metrics would otherwise stop products from adopting the previous NaNExpiry behavior wholesale. The definitions also provide a similar "define your metrics before you use them" approach that's used in Prometheus clients.

More research needs to be done on how to best handle definitions for sub-libraries using go-metrics which expect their configuration from their hosting application. One path offered by this PR is to declare definitions inside the library and have the hosting application "reach into" the lib and append its definitions to the application's when it creates its `PrometheusSink`. A alternative approach has been proposed which is compatible with this definitions PR, and was one of the early iterations on this bug fix: allow prometheus metrics to be "defined" at runtime with additional methods for each metric type in go-metrics. For one example, `InitCounterWithLabels`.

### DIFF
--- a/prometheus/prometheus.go
+++ b/prometheus/prometheus.go
@@ -199,7 +199,6 @@ func initGauges(m *sync.Map, gauges []GaugeDefinition) {
 			Help:        g.Help,
 			ConstLabels: prometheusLabels(g.ConstLabels),
 		})
-		pG.Set(float64(0)) // Initialize at zero
 		m.Store(hash, &gauge{ Gauge: pG })
 	}
 	return
@@ -215,7 +214,6 @@ func initSummaries(m *sync.Map, summaries []SummaryDefinition) {
 			ConstLabels: prometheusLabels(s.ConstLabels),
 			Objectives:  map[float64]float64{0.5: 0.05, 0.9: 0.01, 0.99: 0.001},
 		})
-		pS.Observe(float64(0)) // Initialize at zero
 		m.Store(hash, &summary{ Summary: pS })
 	}
 	return
@@ -229,7 +227,6 @@ func initCounters(m *sync.Map, counters []CounterDefinition) {
 			Help:        c.Help,
 			ConstLabels: prometheusLabels(c.ConstLabels),
 		})
-		pC.Add(float64(0)) // Initialize at zero
 		m.Store(hash, &counter{ Counter: pC })
 	}
 	return
--- a/prometheus/prometheus.go
+++ b/prometheus/prometheus.go
@@ -211,7 +211,9 @@ func initSummaries(m *sync.Map, summaries []SummaryDefinition) {
 		pS := prometheus.NewSummary(prometheus.SummaryOpts{
 			Name:        key,
 			Help:        s.Help,
+			MaxAge:      10 * time.Second,
 			ConstLabels: prometheusLabels(s.ConstLabels),
+			Objectives:  map[float64]float64{0.5: 0.05, 0.9: 0.01, 0.99: 0.001},
 		})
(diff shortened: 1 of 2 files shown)


prometheus/prometheus.go
```diff
@@ -39,19 +60,44 @@ type PrometheusSink struct {
...
+
+// SummaryDefinition can be provided to PrometheusOpts to declare a constant summary that is not deleted on expiry.
+type SummaryDefinition struct {
+	Name []string
+	ConstLabels []metrics.Label
+	Help string
 }
 
-type PrometheusSummary struct {
+type summary struct {
```
cgbaker, line 84: these classes were exported before (even though they didn't need to be), but i don't find any indication that they're used anywhere, so 👍

prometheus/prometheus.go
```diff
@@ -90,43 +140,104 @@ func (p *PrometheusSink) Collect(c chan<- prometheus.Metric) {
...
 
+func initGauges(m *sync.Map, gauges []GaugeDefinition) {
+	for _, g := range gauges {
+		key, hash := flattenKey(g.Name, g.ConstLabels)
+		pG := prometheus.NewGauge(prometheus.GaugeOpts{
+			Name:        key,
+			Help:        g.Help,
+			ConstLabels: prometheusLabels(g.ConstLabels),
+		})
+		pG.Set(float64(0)) // Initialize at zero
```
cgbaker, line 202: should gauges and summaries (which haven't been observed yet) be initialized to zero or `NaN` (as they do after expiration)?

cgbaker (approved):
will go ahead and approve; one question about initializing constant gauges and summaries.

prometheus/prometheus.go
mkcp: Yeah we looked over it and it seemed pretty safe to unexport these because they're only used internally.

prometheus/prometheus.go
mkcp: That is a good question - the initialization part here is a little redundant for counters and gauges because they start at zero in the go prometheus lib anyway. However, initializing summaries to 0 does create an artifact when the process starts and before they're emitted, where the line sits on the 0 like it's a counter or gauge. It's going to be better to delete all of these inits and just use the go prometheus client's defaults.

PR #121, opened 2020-11-25
Export emitRuntimeStats to provide a workaround for goroutine leak.

ncabatoff:
When EnableRuntimeMetrics is true, New will leak a goroutine that collects stats periodically.  Export the method that does the stats collection so that those who are willing to call it themselves can avoid that goroutine leak.


dnephin (approved):
Nice! I've been wanting to fix this leak. This PR looks like a good way to solve the problem.

jefferai (approved):

jefferai:
Doesn't seem like it's a leak if you've enabled the option :-) But happy to have the change.

PR #122, opened 2020-12-29
Use the same Help for metrics with and without labels

pierresouchay:
When metrics are pre-registered with Help and without labels,
when issuing metrics with labels, the help was filled with
the key_name.

But prometheus client enforces help to be consistent accross
several elements, so, it can break at Runtime if a metric
has been declared without its labels and later issued this metric
with labels.

In order to fix this, we store the help declared at startup for
each key, so, when declaring a metric with some labels, we try
to re-used pre-declared help for the metric.

It should fix the issue that appeared with Consul 1.9.x and
fix https://github.com/hashicorp/consul/issues/9471

### DIFF
--- a/prometheus/prometheus.go
+++ b/prometheus/prometheus.go
@@ -58,13 +58,14 @@ type PrometheusSink struct {
 	summaries  sync.Map
 	counters   sync.Map
 	expiration time.Duration
+	help       map[string]string
 }
 
 // GaugeDefinition can be provided to PrometheusOpts to declare a constant gauge that is not deleted on expiry.
 type GaugeDefinition struct {
-	Name []string
+	Name        []string
 	ConstLabels []metrics.Label
-	Help string
+	Help        string
 }
 
 type gauge struct {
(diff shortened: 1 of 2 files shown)


pierresouchay:
Hello @mkcp

Maybe you could have a look at this PR to fix #9303 and #9198

Thank you!

mkcp:
@pierresouchay Awesome!! Thank you for putting together a patch to address hashicorp/consul#9303. I've loaded this up on a local docker cluster and I can confirm that the warning messages at Consul startup and runtime are gone. Also, looking at metrics with dynamic labels - they are correctly batched with the correct metric name & help text. E.g. the `path` label here for consul_api_http.

Screen Shot 2021-01-05 at 3 06 44 PM

mkcp (approved):
Looks great to me! Appreciate the tests as well

pierresouchay:
@mkcp So, we just need a Review + Merge + Tag/Release before submitting the patch in Consul, right?

@dnephin,  or @banks maybe?

banks:
LGTM, thanks Pierre!

banks:
I created [v0.3.6](https://github.com/armon/go-metrics/releases/tag/v0.3.6)

pierresouchay:
Thank you @banks !

Made the update in https://github.com/hashicorp/consul/pull/9510

PR #129, opened 2021-10-12
actually describe PrometheusSink descriptors

FFMMM:
### Overview

Stumbled across this as I was trying to instantiate two `TestAgent`s in consul. See https://github.com/hashicorp/consul/issues/11273 for more context around that papercut.

It all comes down to registering more than on PrometheusSink on a registerer.

Today, that's not possible since we don't actually implement [`.Describe()`](https://github.com/armon/go-metrics/blob/master/prometheus/prometheus.go#L129-L134).

This approach runs over all the gauges, summaries and counters and sends `.Desc()` to the chan.

It also adds a "dummy" value to register "empty" sinks but it relies on a new field: `PrometheusSink.name`.

### Testing
* unit


markan (approved):
Generally looks good. My biggest concern is any back compatibility issues, but this looks pretty safe.

prometheus/prometheus.go
```diff
@@ -126,11 +134,41 @@ func NewPrometheusSinkFrom(opts PrometheusOpts) (*PrometheusSink, error) {
...
+		}
+		co := value.(*counter)
+		co.Describe(c)
+		return true
+	})
+
+	// dummy value to be able to register and unregister "empty" sinks
+	// Note this is not actually retained in the PrometheusSink so this has no side effects
+	// on the caller's sink. So it shouldn't show up to any of its consumers.
+	prometheus.NewGauge(prometheus.GaugeOpts{Name: p.name, Help: p.name}).Describe(c)
```
dnephin, line 170: I know this already exists, but I am surprised it is necessary because those same docs say this:

	// Sending no descriptor at all marks the Collector as “unchecked”,
	// i.e. no checks will be performed at registration time, and the
	// Collector may yield any Metric it sees fit in its Collect method.
	
That to me suggests that it would be safe to send nothing on the channel, but maybe I am interpreting it wrong.

FFMMM:
> We will still have some ephemeral metrics that are not registered when the process starts. When those metrics are emitted our Describe method will start returning different descriptions. I'm not sure what problem that will cause, but it seems like the implementation in this PR might violate that requirement.

Yes, that's spot on -- thanks for the call out! 

So funcs like:

```
...
func (p *PrometheusSink) SetGaugeWithLabels(parts []string, val float32, labels []metrics.Label) {
...
func (p *PrometheusSink) AddSampleWithLabels(parts []string, val float32, labels []metrics.Label) {
...
func (p *PrometheusSink) IncrCounterWithLabels(parts []string, val float32, labels []metrics.Label) {
...
```

store metrics outside of the respective `init*` counterparts. If we call `.Describe()` before any of the funcs above are called (and we do call them over in consul land), the set of the return values will be different than when we call `.Describe()` after those funcs were called.

I'm interpreting this behavior in two ways:
* it is consistent with the spec -- as long as the `Collector` is the same, `Describe()` idempotently returns a super set of the metrics. Since we modify/ mutate the `Collector` metrics in the funcs above, for these new mutations, the `.Describe()` impl still respects the spec

or

* it is inconsistent with the spec -- if we call at T_0 and T_1, T_1 > T_0, there's a chance that `.Describe()` may return different super sets;, a very non idempotent behavior, regardless of the fact that the `Collector` metrics may have changed.

---

#### Pragmatic Hat

As far as I can see, we only **really** care about `.Describe()` when registering/ unregistering a `.Collector`, although others may relay more on the func, hence the need to make a change that doesn't break existing behavior.

Maybe then we could go with @markan's suggestion and just rely on the `Name`:

```
func (p *PrometheusSink) Describe(c chan<- *prometheus.Desc) {
	prometheus.NewGauge(prometheus.GaugeOpts{Name: p.name, Help: p.name}).Describe(c)
}
```

or something else that uniquely identifies the `Collector` and will not change as we continue to store metrics.

prometheus/prometheus.go
FFMMM: Thanks for the pointers -- I'll update this PR or open a new one that has just the `.Name` change.

> any ephemeral metric will change the list of descriptors, which will make it impossible to deregister

hm, just to be explicit -- I think there can be a race between deregistering a `Collector` type and modifying its underlying metrics that can make the `Unregister` fail. It wouldn't be a deterministic problem, but a racy one.

---

I took `superset` to mean the set of all possible metrics that a `Collector` collects and more. Where "more" can be the addition of the `p.Name` in the current form of the code or any other descriptor.

prometheus/prometheus_test.go
```diff
@@ -54,6 +54,64 @@ func TestNewPrometheusSink(t *testing.T) {
...
+
+	reg := prometheus.DefaultRegisterer
+
+	if reg == nil {
+		t.Fatalf("Expected default register to be non nil, got nil.")
+	}
+
+	if err != nil {
+		t.Fatalf("err = %v, want nil", err)
+	}
```
dnephin, line 82: Minor style recommendation: generally errors should be checked immediately after the line the where they are assigned, with no extra newline between the statements. That formatting is encoded in https://github.com/mvdan/gofumpt as "No empty lines before a simple error check".



```suggestion
	sink1, err := NewPrometheusSinkFrom(cfg)
	if err != nil {
		t.Fatalf("err = %v, want nil", err)
	}
	
	reg := prometheus.DefaultRegisterer
	if reg == nil {
		t.Fatalf("Expected default register to be non nil, got nil.")
	}
```

prometheus/prometheus_test.go
```diff
@@ -54,6 +54,64 @@ func TestNewPrometheusSink(t *testing.T) {
...
+		Expiration:         15 * time.Second,
+		GaugeDefinitions:   append([]GaugeDefinition{}, gaugeDef2),
+		SummaryDefinitions: append([]SummaryDefinition{}),
+		CounterDefinitions: append([]CounterDefinition{}),
+		// commenting out the name to point out that the default name will be used here instead
+		// Name:               "sink2",
+	}
+
+	sink2, err := NewPrometheusSinkFrom(cfg2)
+
```
dnephin, line 99:
```suggestion
```

PR #132, opened 2022-04-14
Adds Shutdown to Metrics API

ggambetti:
Short lived applications run the risk of losing metrics that are
generated near the end of their lifetimes because MetricSinks can
and do buffer data locally. The exact amount of data loss depends on
MetricSink implementation, caller configuration, and the timing of
the last metric sync.

This adds a Shutdown function to the package API and the MetricSink
interface. Shutdown flushes locally buffered data to storage, and
then frees resources allocated to the MetricSink.

Currently not all MetricSinks support exiting, and so resource
release when calling Shutdown is best effort. Since Shutdown is
intended for use immediately prior to application exit this is deemed
acceptable, since resource leakage is minimized in this case.

### DIFF
--- a/const_unix.go
+++ b/const_unix.go
@@ -1,3 +1,4 @@
+//go:build !windows
 // +build !windows
 
 package metrics
--- a/const_unix.go
+++ b/const_unix.go
@@ -1,4 +1,3 @@
-//go:build !windows
 // +build !windows
 
 package metrics
--- a/datadog/dogstatsd.go
+++ b/datadog/dogstatsd.go
@@ -120,6 +120,7 @@ func (s *DogStatsdSink) AddSampleWithLabels(key []string, val float32, labels []
 	s.client.TimeInMilliseconds(flatKey, float64(val), tags, rate)
 }
 
+// Shutdown disables further metric collection, blocks to flush data, and tears down the sink.
 func (s *DogStatsdSink) Shutdown() {
 	s.client.Close()
 }
--- a/datadog/dogstatsd.go
+++ b/datadog/dogstatsd.go
@@ -120,6 +120,10 @@ func (s *DogStatsdSink) AddSampleWithLabels(key []string, val float32, labels []
 	s.client.TimeInMilliseconds(flatKey, float64(val), tags, rate)
 }
 
+func (s *DogStatsdSink) Shutdown() {
+	s.client.Close()
+}
+
 func (s *DogStatsdSink) getFlatkeyAndCombinedLabels(key []string, labels []metrics.Label) (string, []string) {
 	key, parsedLabels := s.parseKey(key)
 	flatKey := s.flattenKey(key)
--- a/start.go
+++ b/start.go
@@ -144,3 +144,13 @@ func UpdateFilter(allow, block []string) {
 func UpdateFilterAndLabels(allow, block, allowedLabels, blockedLabels []string) {
 	globalMetrics.Load().(*Metrics).UpdateFilterAndLabels(allow, block, allowedLabels, blockedLabels)
 }
+
+// Shutdown flushes and disables metric collection, blocking while waiting for this to complete.
+// WARNING: Not all MetricSink backends support this functionality, and calling this will cause resource leaks.
+// This is intended for use immediately prior to application exit.
+func Shutdown() {
+	m := globalMetrics.Load().(*Metrics)
+  // Replace global metrics with the BlackholeSink like how init setup the library.
+	globalMetrics.Store(&Metrics{sink: &BlackholeSink{}})
+	m.Shutdown()
+}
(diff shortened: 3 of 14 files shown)


datadog/dogstatsd.go
```diff
@@ -120,6 +120,10 @@ func (s *DogStatsdSink) AddSampleWithLabels(key []string, val float32, labels []
 	s.client.TimeInMilliseconds(flatKey, float64(val), tags, rate)
 }
 
+func (s *DogStatsdSink) Shutdown() {
+	s.client.Close()
```
FFMMM, line 124: I may be wrong here, but this too can block https://github.com/DataDog/datadog-go/blob/v5.1.0/statsd/statsd.go#L698

This is fine, I'd just add docstrings around these new APIs and call that out.

start.go
```diff
@@ -144,3 +144,13 @@ func UpdateFilter(allow, block []string) {
...
 	globalMetrics.Load().(*Metrics).UpdateFilterAndLabels(allow, block, allowedLabels, blockedLabels)
 }
+
+// Shutdown flushes and disables metric collection, blocking while waiting for this to complete.
+// WARNING: Not all MetricSink backends support this functionality, and calling this will cause resource leaks.
+// This is intended for use immediately prior to application exit.
+func Shutdown() {
+	m := globalMetrics.Load().(*Metrics)
+	// Replace global metrics with the BlackholeSink like how init setup the library.
+	globalMetrics.Store(&Metrics{sink: &BlackholeSink{}})
```
FFMMM, line 154: I don't understand this line 100% <-- would you mind helping me understand why we need to do this?

start.go
```diff
@@ -144,3 +144,13 @@ func UpdateFilter(allow, block []string) {
 func UpdateFilterAndLabels(allow, block, allowedLabels, blockedLabels []string) {
 	globalMetrics.Load().(*Metrics).UpdateFilterAndLabels(allow, block, allowedLabels, blockedLabels)
 }
+
+// Shutdown flushes and disables metric collection, blocking while waiting for this to complete.
+// WARNING: Not all MetricSink backends support this functionality, and calling this will cause resource leaks.
+// This is intended for use immediately prior to application exit.
+func Shutdown() {
```
FFMMM, line 151: Could we add a `Test_GlobalMetrics_Shutdown` test in `start_test.go` for a simple smoke screen?

FFMMM (commented):
Thanks for opening this 💯 It looks good; I took a level 1 pass and added some questions and comments.

start.go
ggambetti: The goal is to explicitly break the following example, so that library consumers build their applications knowing that nothing is collected post `Shutdown`, and don't accidentally become dependant on collecting metrics post-`Shutdown`.
```
import "github.com/armon/go-metrics"

func main() {
  // The implementation of the metrics library means
  // that {"hello", "world"} will only ever have the value 1.
  metrics.New(...)
  metrics.IncrCounter([]string{"hello", "world"}, 1)
  metrics.Shutdown()
  metrics.IncrCounter([]string{"hello", "world"}, 1)
}
```

Resetting the library to the `BlackholeSink` isn't strictly necessary. But between applications not cleaning up goroutines (that generate calls to the library), and `MetricSink` implementations not supporting shutdown/close, it's possible that applications collect metrics and then upload them in the small window between `Shutdown` and exit. This makes it pretty explicit that this isn't intended or supported.

There is a race condition here: callers getting a pointer to the underlying `Metrics` struct and then completing their call into the `MetricSink` after `Shutdown` finishes. `MetricSink` must be thread-safe, so this is likely fine.

const_unix.go
```diff
@@ -1,3 +1,4 @@
+//go:build !windows
```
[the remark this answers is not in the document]
ggambetti: I've removed them as they are the only edits to those files.

start.go
FFMMM: Thanks for laying out for me here 💯 I see now the issue that could happen; I'll resolve this comment for now.

FFMMM (approved):
hey @ggambetti this is great! Thanks for addressing all the my comments 💯 🤝 

I think this is ready to merge 🚀 . AFAIK, @banks is the person I know that can help merge and maybe cut a release for these changes 🙈 . I hope I'm not volunteering Paul here.

banks (approved):
Thanks @ggambetti this looks like a clean solution to this issue!

heucuva:
Was there a reason why this was a backwards compatibility breaking change with only a [patch-level](https://semver.org/) version increase (i.e: v0.3.10 -> v0.3.11)? Was a little caught off guard by it.

banks:
@heucuva Apologies, that's my fault. Since this was purely additive I didn't think it was a breaking change. I missed that since it was a public interface addition, other callers implementing their own interfaces would be impacted. Honestly, even without the interface change this probably should have been a minor version at least. My mistake.

This repo is in the process of being moved to an official team rather than a handful of us and Armon who happen to have commit access. Hopefully that will help with a more thoughtful release processes.

I'm going to talk with a few folks and decide what the best path forward is - perhaps re-release as a new major version and remove the 0.3.11 release from GH.

PR #133, opened 2022-05-11
stage retract for v0.3.11

acpana:
Not really sure if we have anyone using `go` < 1.16 today. This won't work for them since the `retract` directive was introduced in `1.16` 🙈 I think.


banks:
Thanks for this Alex, I took it on in #134.

PR #134, opened 2022-05-24
Retract v0.3.11 since it contains breaking changes.

banks:
This is based on #133 thanks @acpana!

Also bumps supported go versions since retract requires Go 1.16.


acpana (approved):
This LGTM! Let's also call out that we dropped support for 13/14/15 starting with `0.4.0` if that's ok 🙈

banks:
> Let's also call out that we dropped support for 13/14/15 starting with 0.4.0 if that's ok 

great call! Thanks Alex

PR #140, opened 2022-10-18
bump dependencies to fix go 1.18+

bernardoVale:
Bumping prometheus client dependency so we can use a version of[ json-iterator](https://github.com/json-iterator/go/releases/tag/v1.1.12) that does not crash with go 1.18+


nirb-s1:
Any update about this PR, this CVE still part of the last version.

PR #146, opened 2023-01-26
prometheus: prevent panic when incrmenting counter

lgfa29:
The Prometheus [Counter.Add()](https://pkg.go.dev/github.com/prometheus/client_golang/prometheus#Counter) method panics if the value being added is negative. Even if care is taken by consumers to never pass a negative value the panic could still happen due to float conversion or overflow.

This change prevents go-metrics from causing consumers to crash.

A panic like this has been reported in https://github.com/hashicorp/nomad/issues/15861, but, as far as I can tell, Nomad only calls `IncrCounter` and `IncrCounterWithLabels` using positive values.

### DIFF
--- a/prometheus/prometheus.go
+++ b/prometheus/prometheus.go
@@ -364,6 +364,13 @@ func (p *PrometheusSink) IncrCounterWithLabels(parts []string, val float32, labe
 	key, hash := flattenKey(parts, labels)
 	pc, ok := p.counters.Load(hash)
 
+	// Prometheus Counter.Add() panics if val < 0. We don't want this to
+	// cause applications to crash, so log an error instead.
+	if val < 0 {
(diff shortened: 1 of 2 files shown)


tgross (approved):
LGTM

tgross:
@lgfa29 as an alternative because this seems like such a weird case and a programmer error rather than a runtime error, what would you think about shipping a version of Nomad pinned to this branch to see if we can get reports on the error to fix the underlying cause?

lgfa29:
Ah good idea. I will open a PR with this dependency update on Nomad tomorrow. I will probably rebase this as branch to `v0.4.1` to make sure this is the only code change.

banks (approved):
Thanks for confirming this can occur in unexpected ways especially since we are working with floats here!

I think if we've had real-world pain caused, it makes sense to be more defensive here.

PR #148, opened 2023-02-06
Go sched scheduler latency

dhiaayachi:
This adds the scheduler queue latency metric to go-metric. This metric is interesting when profiling a go program.

As go-metrics don't support histograms and this metric is exposed as histogram in go runtime library, I chose to expose it in the following form `bucket*count`. This is not ideal as it's factor of total count of active goroutine so it could be misleading in some cases. That said, total count of goroutines is also available as a different metric, so it could help disambiguous this. 

Also, I think this is sufficient to measure the overall scheduling pressure caused by CPU bottlenecks or bugs in a given go program.

### DIFF
--- a/metrics.go
+++ b/metrics.go
@@ -282,8 +282,9 @@ func (m *Metrics) EmitRuntimeStats() {
 
 func (m *Metrics) readSchedulerLatencyAll() {
 	sample := make([]metrics.Sample, 1)
-	const metricName = "/sched/latencies:seconds"
-	sample[0].Name = metricName
+	const metricName = "sched_latencies_seconds"
+	const goName = "/sched/latencies:seconds"
+	sample[0].Name = goName
 
 	metrics.Read(sample)
 	var signal *metrics.Float64Histogram
@@ -300,7 +301,7 @@ func (m *Metrics) readSchedulerLatencyAll() {
 				bucket = math.MaxFloat32
 			}
 			s := fmt.Sprintf("%f", bucket)
-			m.AddSampleWithLabels([]string{"runtime", strings.ReplaceAll(metricName, "/", "_")}, float32(c)*float32(bucket), []Label{{"bucket", s}})
+			m.AddSampleWithLabels([]string{"runtime", metricName}, float32(c)*float32(bucket), []Label{{"bucket", s}})
 		}
 	}
 }
--- a/metrics.go
+++ b/metrics.go
@@ -1,7 +1,10 @@
 package metrics
 
 import (
+	"fmt"
+	"math"
 	"runtime"
+	"runtime/metrics"
 	"strings"
 	"time"
 
@@ -273,6 +276,33 @@ func (m *Metrics) EmitRuntimeStats() {
 		m.AddSample([]string{"runtime", "gc_pause_ns"}, float32(pause))
 	}
 	m.lastNumGC = num
+
+	m.readSchedulerLatencyAll()
+}
+
+func (m *Metrics) readSchedulerLatencyAll() {
+	sample := make([]metrics.Sample, 1)
+	const metricName = "/sched/latencies:seconds"
+	sample[0].Name = metricName
+
+	metrics.Read(sample)
+	var signal *metrics.Float64Histogram
+	if len(sample) >= 1 && sample[0].Value.Kind() == metrics.KindFloat64Histogram {
+		signal = sample[0].Value.Float64Histogram()
+	}
+	if signal != nil {
+		for i, c := range signal.Counts {
+			bucket := signal.Buckets[i]
+			if math.IsInf(bucket, -1) {
+				bucket = 0
+			}
+			if math.IsInf(bucket, 1) {
+				bucket = math.MaxFloat32
+			}
+			s := fmt.Sprintf("%f", bucket)
+			m.AddSampleWithLabels([]string{"runtime", strings.ReplaceAll(metricName, "/", "_")}, float32(c)*float32(bucket), []Label{{"bucket", s}})
+		}
+	}
 }
 
 // Creates a new slice with the provided string value as the first element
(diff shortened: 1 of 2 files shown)


dhiaayachi:
@banks when you have time can you please check this? It would be nice to add this in Consul for perf analysis purpose.

metrics.go
```diff
@@ -273,6 +276,34 @@ func (m *Metrics) EmitRuntimeStats() {
...
+		for i, c := range signal.Counts {
+			bucket := signal.Buckets[i]
+			if math.IsInf(bucket, -1) {
+				bucket = 0
+			}
+			if math.IsInf(bucket, 1) {
+				bucket = math.MaxFloat32
+			}
+			s := fmt.Sprintf("%f", bucket)
+			m.AddSampleWithLabels([]string{"runtime", metricName}, float32(c)*float32(bucket), []Label{{"bucket", s}})
```
banks, line 304: Hmmm. I'm not super sure this is the best way to expose this. I don't know if it is easy to interpret or even meaningful for all sinks.

Samples are generally aggregated either in the library (e.g. Prometheus Summary type) or in the collector (e.g. statsd-like sinks) and then just specific summary metrics emitted to the actual storage system like 50 and 99 percentiles.

The logic here is taking a histogram of a distribution and then emitting a "sample" for each bucket which is the bucket bound (not sure if that's upper, lower or middle?) times the number of observations in that bucket. Since it will do this for each bucket, we'll end up with a set of samples each time this is called (once per interval?) which will then have percentiles taken of those....

Basically I'm not sure what actual measurement you end with in the end or how much to trust it as this process doesn't have a clear statisitical meaning as far as I can tell.

I also don't think this is related to "the number of goroutines" because a single goroutine could be blocked and then become ready thousands of times in a single metrics interval and so therefore would have thousands of scheduling latency samples, yet 10s of thousands of other goroutines might be blocked the whole time and so have zero samples in that period.

In general I'd love to expose this, but perhaps we need to see if we can find a way to do it that is more mathematically significant?

One option would be to specifically emit single gauges here that correspond to the current 50%ile, 99%ile max etc. That would make the equivalent to the output that we get from statsd samples or Prometheus Summaries. Those have downsides (can't aggregate them etc.) but those are the same downsides as every other timing metric in go-metrics today.

The one other downside of doing that here that is different is that we'd be responsible for picking a naming convention - right now each sink/agent does that separately. That means that if we pick the same convention as prometheus say then Prometheus users will see metrics consistent with other timing information but all other sink users will see metrics presented differently to their other timings.

Another option would be that we do similarly to here but we emit a sample for every actual underlying sample e.g. if bucket count is `1000` we loop and emit 1000 separate samples with the mid point of that bucket as the "value". This would preserve the shape of the overall histogram and keep the units and semantics of any percentiles calculated from the samples the same as the ones from the underlying distribution, but it could be a lot more processing to do on a busy server! If that is a concern in testing, we could sample them (e.g. divide count by 100 or something and emit only that many for each) though that would quite possible skew the long tail which is the part we care most about 🤔.

I think emitting gauges with the summary statistics is the most meaningful thing but I could be missing how this code as proposed is equivalent to something meaningful statistically - I'd need to work through examples to feel confident that it preserves important properties of the actual distribution in a way that's reasonable to draw conclusions from. Metrics you can't trust or understand are in some ways worse than no metrics!

banks: The other option would be to emit in the same format prometheus emits histograms - as cumulative bucket gauges.

The label for each is `le=bucketUpperBound`. This would allow prometheus compatible backends at least to calculate and aggregate arbitrary percentiles or statistics over the results, but it results in num_buckets new distinct series which I'm guessing is way to many. So we could do equivalent of Prometheus Histrograms on client side and define boundaries we care about like SLOs and aggregate the buckets around those, e.g. `< 0.1ms, <1ms, <5ms, <20ms` Any significant activity over 20ms is basically extreme scheduler latency anyway?

banks: This is how the buckets are calculated FWIW although this is a detail that may change not a stable API we should assume: https://go.dev/src/runtime/histogram.go?s=4596:4624

This is what they look like visualised (although I can't get go playground to do enough work to actually see scheduling delays).

https://play.golang.com/p/IP45yXPDib4

That's clearly too many buckets just to emit, so I thing the options are between emitting a handful of carefully chosen buckets as gauges, or emitting some carefully chosen percentiles.

Prom docs have a good summary of the pros and cons so I'll just link there rather than repeat: https://prometheus.io/docs/practices/histograms/#quantiles

What do you think Dhia?

dhiaayachi: TBH I struggled a lot with how to present those metrics in the library as histograms are not natively supported in go-metrics. The reason I went with this approach, and I'm not trying to say we should keep it, is that I needed some form of aggregate that would give me an idea of how much overall latency is happening in the scheduler queue. I tried multiple approaches and each have it's limitation:

- calculating an average latency would need more than 1 sample and is usually done at the statd or prometheus level and would be difficult to make as we need to have more than 1 sample
- only emitting a sample around the bucket with the max number of samples would be not really representative as some routine could have high latency but another routine is rescheduled fast enough to fill another smaller bucket
- emit all the buckets as label, it make processing too expensive at the aggregation level and grafana would choke on the number of labels.

> That's clearly too many buckets just to emit, so I thing the options are between emitting a handful of carefully chosen buckets as gauges, or emitting some carefully chosen percentiles.

I like the idea of emitting a number of chosen bucket, but I think it won't work as if the buckets vary enough in time we will end up with all the buckets available at the aggregation level, which is not ideal.

Maybe emitting p99, p90 and p50 percentiles is more reasonable even if it would have limited usage 🤔

dhiaayachi: I was referring to the complete picture but yes we should be covered for most of the cases. I will try to work on a version that expose those when I have some time.

dhiaayachi:
Closing this as it need more work to get the right metrics exposed!

PR #151, opened 2023-04-03
Add GaugeFloat64 API

biazmoreira:
This adds a new API to send and store gauge metrics as float64s. Previously, they were being stored as float32s which isn't enough precision for storing unix timestamps. 

Related issue: https://github.com/hashicorp/vault/issues/19594 and https://github.com/hashicorp/go-metrics/issues/150

### DIFF
--- a/metrics.go
+++ b/metrics.go
@@ -42,6 +42,40 @@ func (m *Metrics) SetGaugeWithLabels(key []string, val float32, labels []Label) 
 	m.sink.SetGaugeWithLabels(key, val, labelsFiltered)
 }
 
+func (m *Metrics) SetPrecisionGauge(key []string, val float64) {
+	m.SetPrecisionGaugeWithLabels(key, val, nil)
+}
+
+func (m *Metrics) SetPrecisionGaugeWithLabels(key []string, val float64, labels []Label) {
+	if m.HostName != "" {
+		if m.EnableHostnameLabel {
+			labels = append(labels, Label{"host", m.HostName})
+		} else if m.EnableHostname {
+			key = insert(0, m.HostName, key)
+		}
+	}
+	if m.EnableTypePrefix {
+		key = insert(0, "gauge", key)
+	}
+	if m.ServiceName != "" {
+		if m.EnableServiceLabel {
+			labels = append(labels, Label{"service", m.ServiceName})
+		} else {
+			key = insert(0, m.ServiceName, key)
+		}
+	}
+	allowed, labelsFiltered := m.allowMetric(key, labels)
+	if !allowed {
+		return
+	}
+	sink, ok := m.sink.(PrecisionGaugeMetricSink)
+	if !ok {
+		// Sink does not implement PrecisionGaugeMetricSink.
+	} else {
+		sink.SetPrecisionGaugeWithLabels(key, val, labelsFiltered)
+	}
+}
+
 func (m *Metrics) EmitKey(key []string, val float32) {
 	if m.EnableTypePrefix {
 		key = insert(0, "kv", key)
--- a/sink.go
+++ b/sink.go
@@ -24,6 +24,12 @@ type MetricSink interface {
 	AddSampleWithLabels(key []string, val float32, labels []Label)
 }
 
+// PrecisionGaugeMetricSink interfae is used to support 64 bit precisions for Sinks, if needed.
+type PrecisionGaugeMetricSink interface {
+	SetPrecisionGauge(key []string, val float64)
+	SetPrecisionGaugeWithLabels(key []string, val float64, labels []Label)
+}
+
 type ShutdownSink interface {
 	MetricSink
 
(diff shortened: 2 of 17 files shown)


hsimon-hashicorp:
Fixes https://github.com/hashicorp/vault/issues/19594

sink.go
```diff
@@ -12,6 +12,9 @@ type MetricSink interface {
 	SetGauge(key []string, val float32)
 	SetGaugeWithLabels(key []string, val float32, labels []Label)
 
+	SetGaugeFloat64(key []string, val float64)
+	SetGaugeFloat64WithLabels(key []string, val float64, labels []Label)
+
```
banks, line 17: Adding methods to an existing public interface is technically a breaking change which would require a whole new major version of the library with new imports etc.

That's because any external implementations of this metrics sink will no longer compile after an upgrade until they also implement these new methods.

I think there are two options here:
 1. Make this a v2 change which for go mod requires moving all the code to a new /v2 subdir with associated go.mod changes etc.
 2. Add these to a _different_ optional interface like `Float64Sink` or something (I know, names are hard...) then implement that for any internal sinks you care about. In the place where we call this method on a sink, type check it to see if it implements this interface first and if not do something. The choice on what to do here is somewhat nuanced. The signatures have no error returns for now because metrics are meant to be "fire and forget" and shouldn't hold up application processing, so the choice is either `panic` and document that it is a developer error to allow a sink that doesn't support this new extended interface to be configured in an app that relies on the new methods you added... or you could just auto-downgrade to float32 which at least results in _some_ metric though if the application bothered to use the 64 bit variant it's still likely to be the wrong thing, or you could just drop the gauge on the floor and maybe log it (though that might require plumbing a logger into the library) and if you do choose this path you probably would want to consider rate-limiting the logs because SetGauge could be called really often in an application and would fill logs really fast if every call logs a warning.

What do you think about those options?

banks (commented):
Code wise this looks great Bianca!

Compatibility wise I think sadly we need to jump through some extra hoops to handle the breaking change the the libraries public API one way or the other. Details inline...

sink.go
biazmoreira: I prefer option 2. Do you have any preferences?

banks: 2 seems more pragmatic to me. It's a little more "messy" in that ideally we'd have a single nice interface but IMO this isn't a big enough issue to warrant a v2 with all that entails.

The implicit option 3 would be to ask if this is even worth solving at all and choose to do nothing and live with the limitation of float32 for now. It's always worth asking the question! I can't say how important that is for vault though or what the implications of not fixing this might be so assuming it's worth it, I'd go for 2.

biazmoreira: I did some more thinking and I believe option 2 is also not possible, since it would break FanoutSink, which relies on every Sink implementing the same interface (MetricsSink).

biazmoreira: I think in order to fix the issue we would indeed need a v2 of the library. 😿

sink.go
```diff
@@ -24,6 +24,11 @@ type MetricSink interface {
 	AddSampleWithLabels(key []string, val float32, labels []Label)
 }
 
+type PrecisionGaugeMetricSink interface {
```
banks, line 27: I realise the rest of this library does a less-than-ideal job of documenting interfaces already, but I'd be inclined to add a doc comment to explain the purpose of this additional interface. Could be a one liner, could refer to the history of not originally supporting 64bit floats, could even link to this PR for more info but it's probably useful context to a future reader who wonders why this is different to MetricSink!

I really like the name, but without any comment I might assume "precision gauge" is a specific concept/semantic I wasn't aware of and wonder where to look for more info. Just a simple line or two to the effect that not all sinks may support it but if 64 bit precisions is needed this allows that would save me time looking up commit history etc!

metrics.go
```diff
@@ -42,6 +42,39 @@ func (m *Metrics) SetGaugeWithLabels(key []string, val float32, labels []Label)
...
+			key = insert(0, m.ServiceName, key)
+		}
+	}
+	allowed, labelsFiltered := m.allowMetric(key, labels)
+	if !allowed {
+		return
+	}
+	sink, ok := m.sink.(PrecisionGaugeMetricSink)
+	if !ok {
+		panic("Sink does not implement Float64MetricSink.")
```
banks, line 73:
```suggestion
		panic("Sink does not implement PrecisionGaugeMetricsSink.")
```

start.go
```diff
@@ -105,6 +105,14 @@ func SetGaugeWithLabels(key []string, val float32, labels []Label) {
...
 }
 
+func SetPrecisionGauge(key []string, val float64) {
+	globalMetrics.Load().(*Metrics).SetPrecisionGauge(key, val)
+}
+
+func SetPrecisionGaugeWithLabels(key []string, val float64, labels []Label) {
+	globalMetrics.Load().(*Metrics).SetPrecisionGaugeWithLabels(key, val, labels)
+}
+
```
banks, line 115: This file in general is not well commented but I would lean towards adding doc comments for these functions. Partly because they form part of the libraries core API but specifically because they might panic if used with an incompatible sink and I think callers should be very clear about that before using them!

banks (approved):
Amazing work Bianca!

I've approved because code-wise this all looks good to me.

I left a few suggestions inline mostly around commenting the new behaviour in doc comments to make the library usage and the footguns more obvious for future consumers or code readers! This library is not well commented in general so it's a bit of a drop in the ocean, but especially the new functions that may panic etc. feel like they should be clearly documented to me!

sink.go
```diff
@@ -57,6 +64,20 @@ func (fh FanoutSink) SetGaugeWithLabels(key []string, val float32, labels []Labe
...
+func (fh FanoutSink) SetPrecisionGauge(key []string, val float64) {
+	fh.SetPrecisionGaugeWithLabels(key, val, nil)
+}
+
+func (fh FanoutSink) SetPrecisionGaugeWithLabels(key []string, val float64, labels []Label) {
+	for _, s := range fh {
+		if s64, ok := s.(PrecisionGaugeMetricSink); ok {
+			s64.SetPrecisionGaugeWithLabels(key, val, labels)
+		} else {
+			s.SetGaugeWithLabels(key, float32(val), labels)
```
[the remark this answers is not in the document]
biazmoreira: > We chose to panic in the general case of a sink not supporting rather than silently downsample.

This wasn't a thought-out decision. Do you think it's worth downsampling the values instead of panicking in case the type check is unsuccessful?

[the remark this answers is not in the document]
banks: Yeah I liked the panic option for straightforward case because I think that makes it harder to accidentally use the library incorrectly if your sink doesn't support the feature.

In this case I'd maybe lean towards doing neither panic or downsample and instead just not call anything on an unsupported sink for a precision gauge. Possibly wrong data seems more dangerous or likely to cause problems than missing data - if the fact it's missing is noticed then it might prompt some research into why whereas it just being wrong is much less likely to be spotted and could lead to incorrect analysis downstream by humans or automation.

Whatever option we should add a doc comment to call it out as a footgun the developer is responsible for dealing with.

PR #152, opened 2023-05-01
Update and fix errors `cannot use generic type iradix.Tree[T any] without instantiation` ; `cannot infer T`

kirisakow:
* `go.mod`: upgrade `go 1.12` -> `go 1.20`, update dependencies
* `start.go`, `metrics.go`: fix errors `cannot use generic type iradix.Tree[T any] without instantiation` ; `cannot infer T`


tgross:
Doing a little issue cleanup. We've got no plans to update immutable radix to v2 here.

PR #153, opened 2023-05-05
resources leak fix

artemseleznev:
resources leak fix


grey3228:
@artemseleznev  hi, I would also suggest to close `sock` at the start of CONNECT section before assignment because it may be already initialized and hold descriptor (if I understand it correctly):

```
CONNECT:
	if sock != nil {
		sock.Close()
	}
	// Attempt to connect
	sock, err = net.Dial("tcp", s.addr)
```

PR #155, opened 2023-06-07
Add PrecisionGauge to Circonus

biazmoreira:
Add PrecisionGauge to Circonus

### DIFF
--- a/circonus/circonus_test.go
+++ b/circonus/circonus_test.go
@@ -99,6 +99,34 @@ func TestSetGauge(t *testing.T) {
 	}
 }
 
+func TestSetPrecisionGauge(t *testing.T) {
+	q := make(chan string)
+
+	server := fakeBroker(q)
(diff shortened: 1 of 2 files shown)


circonus/circonus_test.go
```diff
@@ -99,6 +99,34 @@ func TestSetGauge(t *testing.T) {
...
+	cfg := &Config{}
+	cfg.CheckManager.Check.SubmissionURL = server.URL
+
+	cs, err := NewCirconusSink(cfg)
+	if err != nil {
+		t.Errorf("Expected no error, got '%v'", err)
+	}
+
+	go func() {
+		cs.SetPrecisionGauge([]string{"foo", "bar"}, 1)
```
banks, line 117: Is it worth testing this with a value that can't be represented in a float32 to be sure it actually works as expected?

Issue #154, opened 2023-06-06
FanoutSink key modified across sinks when DogStatsdSink is enabled


Achooo:
### Problem
1. FanoutSink does not copy the key before passing it to different sink.
2. DogStatsdSink overrides the key slice in Line 73 when the hostname is the same as a key value:

https://github.com/hashicorp/go-metrics/blob/858601465fcd6976fa4b951419becea272d25579/datadog/dogstatsd.go#L63-L77

### Reproduction Steps

1. Create a FanoutSink
2. Add a DogStatsdSink with a hostname
3. Add another Sink (e.g. InmemSink)
4. Call `SetGaugeWithLabels()` with the hostname as a value in the keys

Example of a test I added in `datadog/dogstatsd_test.go` :
```go
func TestFanoutSink(t *testing.T) {
	server, _ := setupTestServerAndBuffer(t)
	defer server.Close()

	dog, err := NewDogStatsdSink(DogStatsdAddr, "consul")
	require.NoError(t, err)

	inmem := metrics.NewInmemSink(10*time.Second, 10*time.Second)

	fSink := metrics.FanoutSink{dog, inmem}
	fSink.SetGaugeWithLabels([]string{"consul", "metric"}, 10, []metrics.Label{{Name: "a", Value: "b"}})

	intervals := inmem.Data()
	require.Len(t, intervals, 1)

	if intervals[0].Gauges["consul.metric;a=b"].Value != 10 {
		t.Fatalf("bad val: %v", intervals[0].Gauges)
	}
}
```

### Output
```bash
--- FAIL: TestFanoutSink (0.00s)
    /<>/HashiCorp/go-metrics/datadog/dogstatsd_test.go:164: bad val: map[metric.metric;a=b:{metric.metric  10 [{a b}] map[]}]
FAIL
FAIL	github.com/hashicorp/go-metrics/datadog	0.694s
FAIL
```

### Expected Output
I would expect the key in the inmemsink to remain `consul.metric` but it is modified to `metric.metric`

### Fix
The FanoutSink should ideally copy the key before passing it to sinks to avoid modifications

PR #156, opened 2023-06-08
Prevent dogstatsd sink from clobbering the metric key for other sinks

mkeeler:
Fixes: #154 

If using the Fanout sink and including the dogstatsd sink, any sinks sent the metric AFTER dogstatsd could see invalid metric names. This was due to some inadvertent modifications to the original metric key. The new implementation ensure that when modifying the orignial metric key, a new backing array is used and will have key parts copied into it instead of changing elements in the original metric key.

### DIFF
--- a/datadog/dogstatsd.go
+++ b/datadog/dogstatsd.go
@@ -70,7 +70,10 @@ func (s *DogStatsdSink) parseKey(key []string) ([]string, []metrics.Label) {
 	// Splice the hostname out of the key
 	for i, el := range key {
 		if el == hostName {
-			key = append(key[:i], key[i+1:]...)
+			// We need an intermediate key to prevent clobbering the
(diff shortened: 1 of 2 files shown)


chapmanc:
This looks good but should we also make a copy of the key as the fan out passes key? To avoid the sinks from affecting each other?

chapmanc (approved):
Lgtm with one question

mkeeler:
I looked at a few other sinks and this one is the only one that modified the key before converting it to a string.

As metrics get emitted a lot, I figured it would put unnecessary pressure on the GC to copy every time when it _should_ be the responsibility of the sink to not modify.

chapmanc:
That's a good call out. We should probably add a unit test to ensure that keys don't get changed. A new sink or changes to existing sinks could easily resurface this.

PR #163, opened 2024-04-19
Remove test dependency on 3rd party library

woutslakhorst:
While looking at the dependencies of our own project, we tracked one of them coming from hashicorp/go-metrics.
The removed library is not contributing anything other as a helper for a single test. Removing a dependency to a 3rd party library reduces the change injection of malicious code in the future.


tgross:
Cherry-picked to https://github.com/hashicorp/go-metrics/pull/208

PR #166, opened 2024-08-07
[fix] Update dogstatsd client

graysonchao:
Hello! We're using go-metrics over at Square and found that go-metrics had trouble connecting to a dogstatsd forwarder listening on a local abstract Unix socket. After some experimentation, we found that updating the underlying dogstatsd client (to pick up https://github.com/DataDog/datadog-go/pull/113) fixed the problem.

I also went through and updated a few test expectations, adding newlines where needed, so that the tests pass. I'm not sure if this was needed due to the dependency update or not.


tgross:
Hi, doing a little issue/PR cleanup and noticed this one. I've covered this in https://github.com/hashicorp/go-metrics/pull/187 but I've pulled you in as co-author on the commit.

PR #175, opened 2025-04-08
Convert compat package to submodule to remove direct dependency on github.com/armon/go-metrics

hazzik:
I think this way it could provide better package management experience

### DIFF
--- a/compat/go.mod
+++ b/compat/go.mod
@@ -1,0 +1,10 @@
+module github.com/hashicorp/go-metrics/compat
+
+go 1.12
+
+require (
+	github.com/armon/go-metrics v0.4.1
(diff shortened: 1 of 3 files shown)


tgross (changes requested):
Hi @hazzik! The armon/go-metrics situation is intended to be temporary while we migrate the ecosystem of projects. We're still in the process of doing so. I'm inclined to say we should just leave everything as is while @mkeeler finishes up that work, unless you've identified a specific downstream problem the current layout causes?

tgross:
I'm going to close this PR. I'm not sure what the current status of the work @mkeeler is doing as I know there was a good bit of roll-out to do across projects, but I'll leave that up to him to respond.

PR #178, opened 2025-06-04
fix staticcheck lint errors and enable it in pr checks

abhijeetviswa:


inmem.go
```diff
@@ -320,7 +320,7 @@ func (i *InmemSink) getInterval() *IntervalMetrics {
 
 	// Re-check for an existing interval now that the lock is re-acquired.
 	n = len(i.intervals)
-	if n > 0 && i.intervals[n-1].Interval == intv {
+	if n > 0 && i.intervals[n-1].Interval.Equal(intv) {
```
abhijeetviswa, line 323: Fixes:
```
go-metrics/inmem.go:312:14: QF1009: probably want to use time.Time.Equal instead (staticcheck)
        if n > 0 && i.intervals[n-1].Interval == intv {
                    ^
go-metrics/inmem.go:323:14: QF1009: probably want to use time.Time.Equal instead (staticcheck)
        if n > 0 && i.intervals[n-1].Interval == intv {
```

metrics_test.go
abhijeetviswa, line 1: Fixes:
```
go-metrics/metrics_test.go:519:2: SA4006: this value of key is never used (staticcheck)
        key := []string{"thing"}
        ^
go-metrics/metrics_test.go:566:2: SA4006: this value of key is never used (staticcheck)
        key := []string{"thing"}
```

sonamtenzin2 (approved):

PR #185, opened 2025-06-11
fix additional lint errors and increase linter scope

abhijeetviswa:


mukeshjc (approved):

inmem_signal_test.go
```diff
@@ -31,7 +31,10 @@ func TestInmemSignal(t *testing.T) {
 	time.Sleep(15 * time.Millisecond)
 
 	// Send signal!
-	syscall.Kill(os.Getpid(), syscall.SIGUSR1)
+	err := syscall.Kill(os.Getpid(), syscall.SIGUSR1)
+	if err != nil {
+		t.Fatalf("failed to signal process: %s", err)
+	}
```
tgross, line 37: What's the circumstance that would hit this error?

abhijeetviswa: That would be platform dependent. After reading the kill(2) man page for MacOS and Linux, I don't see any obvious way for this syscall to fail. Nevertheless, I believe the test won't negatively impact the test.

inmem.go
```diff
@@ -263,37 +265,11 @@ func (i *InmemSink) Data() []*IntervalMetrics {
...
-	}
-	copyCurrent.Counters = make(map[string]SampledValue, len(current.Counters))
-	for k, v := range current.Counters {
-		copyCurrent.Counters[k] = v.deepCopy()
-	}
-	copyCurrent.Samples = make(map[string]SampledValue, len(current.Samples))
-	for k, v := range current.Samples {
-		copyCurrent.Samples[k] = v.deepCopy()
-	}
+	intervals[n-1] = current.deepCopy()
```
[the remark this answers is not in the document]
dadgar: This section looks good

PR #186, opened 2025-06-13
group dependabot prs

abhijeetviswa:


.github/dependabot.yaml
```diff
@@ -10,10 +10,24 @@ updates:
       interval: "weekly"
```
KaushikiAnand, line 10: Can day: sunday be added here.

KaushikiAnand (approved):

PR #201, opened 2026-03-06
Add OpenTelemetry metrics sink

### LINKED ISSUE #200: Add OpenTelemetry metrics sink
## Problem

go-metrics has no native way to export metrics to OpenTelemetry-compatible backends. Users who adopt OTEL must write their own adapter or maintain a separate metrics pipeline.

## Requirements

**Implement `MetricSink`, `PrecisionGaugeMetricSink`, and `ShutdownSink` interfaces**
Drop-in compatibility with existing go-metrics usage. No changes needed at call sites — users just swap the sink.

**Support multiple configuration modes: OTLP endpoint, custom exporter, custom reader, and bring-your-own MeterProvider**
Different teams have different infrastructure. Some just want to point at a collector endpoint. Others need to plug in a custom exporter (e.g., for a proprietary backend), control the reader (e.g., pull-based collection), or share a MeterProvider across libraries. A single rigid mode would force workarounds.

**Map Gauge, Counter, and Sample to OTEL equivalents; handle EmitKey gracefully**
These are the four go-metrics operations. Gauge, Counter, and Sample have natural OTEL counterparts. EmitKey has no equivalent — it should be dropped without breaking callers, with a warning so users know to migrate.

**Default to exponential histograms with opt-in explicit buckets**
Exponential histograms adapt automatically to any value distribution without requiring users to pre-define bucket boundaries. Explicit buckets should remain available for teams that need deterministic boundaries or compatibility with existing dashboards.

**Allow resource configuration (service name, hostname, custom attributes)**
OTEL backends use resource attributes to identify and group metrics by service. The common case (service name + hostname) should be simple. Full resource override should be available for advanced setups.

**Graceful shutdown with flush**
Metrics sinks often buffer data. Without an explicit shutdown path, in-flight metrics are lost on process exit. The sink should flush pending data and respect a configurable timeout.

**Concurrent-safe for hot-path usage**
go-metrics sinks are called from arbitrary goroutines on every operation. The sink must handle concurrent writes without locks on the hot path to avoid becoming a bottleneck.

arungupta:
## Summary

Closes #200

Adds a new `OTELSink` that exports go-metrics to OpenTelemetry-compatible backends via OTLP/gRPC.

**Configuration modes** (in priority order):
1. `MeterProvider` — bring your own provider (caller retains ownership)
2. `Reader` — custom metric reader
3. `Exporter` — custom exporter wrapped in a PeriodicReader
4. `Endpoint` — simple OTLP/gRPC push (TLS by default, `Insecure: true` for plaintext)

**Features:**
- Explicit bucket and exponential histogram support
- Configurable resource attributes, service name, hostname
- Configurable push interval, shutdown timeout, dial timeout, key separator
- Cardinality warning at configurable threshold (default 10k instruments)
- Negative counter guard (dropped with warning, matching PrometheusSink behavior)
- KeySeparator validation against OTEL instrument naming spec
- One-time warning on `EmitKey` calls (no OTEL equivalent)
- Graceful shutdown with `Shutdown()` and context-aware `ShutdownContext(ctx)`
- Instrument creation error logging (once per unique name)
- Compile-time interface checks for `MetricSink`, `PrecisionGaugeMetricSink`, `ShutdownSink`

**Usage:**
```go
sink, err := otel.NewOTELSink(otel.OTELSinkOpts{
    Endpoint:    "localhost:4317",
    Insecure:    true,
    ServiceName: "my-service",
})
if err != nil {
    log.Fatal(err)
}
defer sink.Shutdown()
metrics.NewGlobal(metrics.DefaultConfig("my-service"), sink)
```

## Testing Done

- 60 tests covering all metric types, label propagation, histogram configurations, resource building, shutdown lifecycle, concurrency safety, edge cases (empty keys, nil labels, special characters, negative values)
- All tests pass with `-race` detector
- End-to-end validated with a gRPC test service → OTEL Collector → Victoria Metrics pipeline, comparing OTEL push and Prometheus scrape pipelines side-by-side


arungupta:
Hi @ssagarverma - gentle reminder for review on this.

tgross:
Does it really make sense to pull in all of OTEL for go-metrics? I'd expect projects that are using OTEL would just pull into OTEL directly. There wasn't any discussion on the issue you opened.

arungupta:
Thanks for the feedback @tgross, fair points on both fronts.

**On the dependency concern:** It's true that OTEL + gRPC is heavier than the existing sinks. I'm happy to restructure this as a separate Go submodule with its own `otel/go.mod`. That keeps the OTEL dependencies completely out of the root module's dependency graph. Consumers would `require github.com/hashicorp/go-metrics/otel v0.1.0` separately, and it can version independently from the core library.

**On whether this belongs in go-metrics at all:** The use case is projects that are deeply integrated with go-metrics and can't easily switch to the OTEL SDK directly. SPIFFE SPIRE is a concrete example. Its entire telemetry layer is built on go-metrics (`telemetry.Sink` is a type alias for `metrics.MetricSink`, every metric call flows through `go-metrics` via `FanoutSink`). Adding OTEL export to SPIRE with this sink is ~50 lines of glue code in SPIRE. Without it, the options are a full rewrite of SPIRE's telemetry layer or a one-off adapter maintained separately. There's an open SPIRE issue for exactly this (spiffe/spire#6544).

More broadly, organizations that have standardized on OTEL, such as mine (LinkedIn), and don't support Prometheus scraping infrastructure (or are moving away from it) have no way to get metrics out of go-metrics-based projects without standing up a Prometheus-to-OTEL pipeline as a shim. A native OTEL sink in go-metrics gives the entire ecosystem a clean path.

Should I go ahead with the submodule approach, or would you prefer this live in a separate repo entirely?

tgross:
Hey @arungupta! I don't think we were all that aware there was much of a third-party "ecosystem" around this library, to be honest. I think you're going to run into build issues due to the [build tags for backcompat](https://github.com/hashicorp/go-metrics?tab=readme-ov-file#backwards-compatibility) at some point.

Let me bring this back to a wider audience internally and report back on that discussion.

tgross:
> Let me bring this back to a wider audience internally and report back on that discussion.

It's been pointed out to me that our products do in fact import OTEL metrics, so my concerns aren't that big a deal. I'll give this a review.

otel/benchmark_test.go
```diff
@@ -0,0 +1,831 @@
+// Copyright IBM Corp. 2013, 2025
```
tgross: Why are we including benchmarks at all? Is this meaningful or just overly-eager LLM doing its slop thing?

otel/otel.go
```diff
@@ -0,0 +1,470 @@
...
+
+// bgCtx is a package-level context.Background() to avoid per-call allocation on the hot path.
+var bgCtx = context.Background()
+
+// Compile-time interface checks
+var (
+	_ metrics.MetricSink              = (*OTELSink)(nil)
+	_ metrics.PrecisionGaugeMetricSink = (*OTELSink)(nil)
+	_ metrics.ShutdownSink            = (*OTELSink)(nil)
+)
```
tgross: Put this in tests.

otel/correctness_test.go
```diff
@@ -0,0 +1,819 @@
+// Copyright IBM Corp. 2013, 2025
```
tgross: What are we really testing in this giant test file that isn't already covered by the other tests? It doesn't improve coverage at all, and it sure feels like we're just testing that sync.Map works, which in turn feels like I'm reviewing the unedited output of an LLM instead of something you've spent the time to review and edit yourself.

otel/options.go
```diff
@@ -0,0 +1,134 @@
...
+	DefaultKeySeparator = "."
+)
+
+// OTELSinkOpts configures the OpenTelemetry metrics sink.
+//
+// Configuration modes (in priority order):
+//  1. MeterProvider: Use a fully configured MeterProvider directly
+//  2. Reader: Use a custom metric.Reader (e.g., PeriodicReader with custom exporter)
+//  3. Exporter: Use a custom metric.Exporter (wrapped in PeriodicReader automatically)
+//  4. Endpoint: Use simple OTLP gRPC push to the specified endpoint
```
tgross: The prioritization of these configuration options means that a user can, say, set `Endpoint` and have it silently ignored if a higher-priority configuration option is used, which could be a footgun. Maybe we want to have a `SinkOptsFromXXX` set of functions to configure from the "high priority" options?

otel/otel_test.go
```diff
@@ -0,0 +1,1190 @@
...
+		for i := range sm.Metrics {
+			if sm.Metrics[i].Name == name {
+				return &sm.Metrics[i]
+			}
+		}
+	}
+	return nil
+}
+
+// ==================== Constructor Tests ====================
```
tgross: Can we ditch all these banners everywhere?

tgross (changes requested):
@arungupta I've left some comments about the options configuration, but this needs a serious human editing pass before we can review it further.

otel/otel_test.go
```diff
@@ -0,0 +1,1348 @@
+// Copyright IBM Corp. 2013, 2025
```
tgross: It's 2026, so any new files should say 2026.

otel/otel_test.go
```diff
@@ -0,0 +1,1348 @@
...
+	}
+	if len(gauge.DataPoints) != 1 {
+		t.Fatalf("expected 1 data point, got %d", len(gauge.DataPoints))
+	}
+	if gauge.DataPoints[0].Value != 100 {
+		t.Errorf("expected value 100, got %f", gauge.DataPoints[0].Value)
+	}
+}
+
+func TestSetGauge_SpecialCharactersInLabelNames(t *testing.T) {
```
tgross: We have a whole bunch of subtests _and_ much of this is covered in `TestSpecialCharactersInLabels`. I was going to recommend crunching this down to a single test, but it's not even clear to me what sink code this actually testing. Isn't all of the behavior being tested here owned by the OTEL libraeries?

otel/otel_test.go
```diff
@@ -0,0 +1,1348 @@
...
+	sink, reader := newTestSink(t)
+	defer sink.Shutdown()
```
tgross:
```suggestion
	t.Cleanup(sink.Shutdown)`
```

(Typical of a lot of the tests)

otel/otel_test.go
```diff
@@ -0,0 +1,1348 @@
...
+			t.Errorf("label %q not found", label.Name)
+			continue
+		}
+		if val.AsString() != label.Value {
+			t.Errorf("label %q: expected %q, got %q", label.Name, label.Value, val.AsString())
+		}
+	}
+}
+
+func TestZeroValues(t *testing.T) {
```
tgross: Again, what is this code actually testing? All these "edge values" tests could easily be consolidation, but I question whether they're actually exercising code you've added in this PR rather than the OTEL SDK.

otel/otel_test.go
```diff
@@ -0,0 +1,1348 @@
...
+		Insecure: true,
+	})
+	if err != nil {
+		t.Fatalf("unexpected error: %v", err)
+	}
+	if sink == nil {
+		t.Fatal("sink should not be nil")
+	}
+	sink.Shutdown()
+}
```
tgross: These two tests plus `TestNewOTELSink_ValidEndpoint` take _30 seconds_ to complete and can't even be parallelized because we're taking up a port. What are you testing here that isn't better covered elsewhere and won't take up 30s of valuable test run time?

otel/otel.go
```diff
@@ -0,0 +1,446 @@
...
+
+	exporter, err := buildExporter(opts)
+	if err != nil {
+		return nil, err
+	}
+
+	pushInterval := defaultDuration(opts.PushInterval, DefaultPushInterval)
+	reader := metric.NewPeriodicReader(exporter, metric.WithInterval(pushInterval))
+
+	// Build the resource.
```
tgross: This is a `i+= // increment i` type of comment. Let's avoid those.

```suggestion
```

otel/options.go
```diff
@@ -0,0 +1,116 @@
...
+
+	// DefaultExpHistogramMaxScale is the maximum scale factor for exponential
+	// histograms. Higher values allow finer-grained bucket boundaries.
+	DefaultExpHistogramMaxScale = 20
+)
+
+// OTELSinkOpts configures the OpenTelemetry metrics sink for endpoint mode.
+// The sink creates an OTLP exporter (gRPC or HTTP), wraps it in a
+// PeriodicReader, and builds a MeterProvider internally. The sink owns the
+// provider and shuts it down when Shutdown is called.
```
tgross: This bit of comment likely fits better on the constructor, not the options. But does it belong there either? It seems like this is "internal to the library" information rather than something someone would need to know as a consumer. But I'm open to being suggestion otherwise.

otel/otel.go
```diff
@@ -0,0 +1,446 @@
...
+	"go.opentelemetry.io/otel/sdk/resource"
+	"google.golang.org/grpc/credentials"
+)
+
+// Compile-time interface checks.
+var (
+	_ metrics.MetricSink               = (*OTELSink)(nil)
+	_ metrics.PrecisionGaugeMetricSink = (*OTELSink)(nil)
+	_ metrics.ShutdownSink             = (*OTELSink)(nil)
+)
```
tgross: Move these to the test file

otel/otel.go
```diff
@@ -0,0 +1,446 @@
...
+// SetPrecisionGauge sets a gauge value with 64-bit precision.
+func (s *OTELSink) SetPrecisionGauge(key []string, val float64) {
+	s.SetPrecisionGaugeWithLabels(key, val, nil)
+}
+
+// SetPrecisionGaugeWithLabels sets a gauge value with 64-bit precision and labels.
+func (s *OTELSink) SetPrecisionGaugeWithLabels(key []string, val float64, labels []metrics.Label) {
+	name := s.flattenKey(key)
+	gauge := s.getOrCreateGauge(name)
+	gauge.Record(context.Background(), val, otelmetric.WithAttributeSet(labelsToAttributes(labels)))
```
tgross: All these `context.Background()` calls for recording means that we can potentially have a bunch of lingering in-flight requests during shutdown. Maybe we should create a context in the constructor that they can all use, which we close in the `Shutdown` method?

otel/otel.go
```diff
@@ -0,0 +1,446 @@
...
+}
+
+// ShutdownContext stops the sink and flushes any remaining metrics.
+// It blocks until metrics are flushed or the context is cancelled.
+// Returns an error if shutdown fails or times out.
+// Only the first call performs shutdown; subsequent calls return the result
+// of the first call regardless of the provided context.
+func (s *OTELSink) ShutdownContext(ctx context.Context) error {
+	s.shutdownOnce.Do(func() {
+		if p, ok := s.provider.(interface{ Shutdown(context.Context) error }); ok {
```
tgross: What's this all about? The OTEL API docs are... a bit of a mess here.

otel/otel.go
```diff
@@ -0,0 +1,446 @@
...
+	})
+	return s.shutdownErr
+}
+
+// flattenKey joins key parts with dots.
+func (s *OTELSink) flattenKey(parts []string) string {
+	return strings.Join(parts, ".")
+}
+
+// getOrCreateGauge returns an existing gauge or creates a new one.
```
tgross: We have three of these very similar get-or-create functions and all of them are used exactly once. This suggests we can inline them and maybe pull out some abstraction between them.

tgross (changes requested):
I've made a first pass review, but the tests still need a lot of work in terms of making them meaningful.

otel/otel.go
arungupta: This was needed for supporting a custom meter provider to be specified by the caller for various kinds of overrides. But since the scope here is limited to go-metrics, I've removed that functionality to keep things simple.

arungupta:
> I've made a first pass review, but the tests still need a lot of work in terms of making them meaningful.

@tgross - Thanks for taking the time. Addressed your comments.

vjdhama:
@arungupta We're looking forward for this OTEL metrics sink to land, because of similar needs.

Are you still actively working on this change?

jzeng4:
This is a feature we need in SPIFFE/SPIRE, which currently uses `go-metrics` for Prometheus metrics. We also need OpenTelemetry metrics support, so this feature would be very helpful for us.

PR #206, opened 2026-06-19
prometheus: periodically clear expired metrics

tgross:
The Prometheus sink clears expired metrics when they're collected. But this means that if a metrics collector fails, metrics will accumulate in the sink and consume potentially large amounts of memory. This has been implicated in a number of reported problems with Nomad as well as internal incidents.

Allow consumers to spin up a background goroutine on the sink that periodically GCs expired metrics if no collection has happened for longer than twice the expiry.

Ref: https://github.com/hashicorp/nomad/issues/27013
Ref: https://github.com/hashicorp/nomad/issues/18113
Ref: https://hashicorp.atlassian.net/browse/NMD-1059
Ref: https://hashicorp.atlassian.net/browse/NMD-555
Ref: `#hashi-inc-8583`

### DIFF
--- a/prometheus/prometheus.go
+++ b/prometheus/prometheus.go
@@ -6,10 +6,12 @@
 package prometheus
 
 import (
+	"context"
 	"fmt"
 	"log"
 	"strings"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/hashicorp/go-metrics"
@@ -56,12 +58,13 @@ type PrometheusOpts struct {
 
 type PrometheusSink struct {
 	// If these will ever be copied, they should be converted to *sync.Map values and initialized appropriately
-	gauges     sync.Map
-	summaries  sync.Map
-	counters   sync.Map
-	expiration time.Duration
-	help       map[string]string
-	name       string
+	gauges         sync.Map
+	summaries      sync.Map
+	counters       sync.Map
+	expiration     time.Duration
+	lastCollection atomic.Int64
+	help           map[string]string
+	name           string
 }
 
 // GaugeDefinition can be provided to PrometheusOpts to declare a constant gauge that is not deleted on expiry.
@@ -116,12 +119,13 @@ func NewPrometheusSinkFrom(opts PrometheusOpts) (*PrometheusSink, error) {
 		name = "default_prometheus_sink"
 	}
 	sink := &PrometheusSink{
-		gauges:     sync.Map{},
-		summaries:  sync.Map{},
-		counters:   sync.Map{},
-		expiration: opts.Expiration,
-		help:       make(map[string]string),
-		name:       name,
+		gauges:         sync.Map{},
+		summaries:      sync.Map{},
+		counters:       sync.Map{},
+		expiration:     opts.Expiration,
+		lastCollection: atomic.Int64{},
+		help:           make(map[string]string),
+		name:           name,
 	}
 
 	initGauges(&sink.gauges, opts.GaugeDefinitions, sink.help)
@@ -157,6 +161,7 @@ func (p *PrometheusSink) Collect(c chan<- prometheus.Metric) {
 // collectAtTime allows internal testing of the expiry based logic here without
 // mocking clocks or making tests timing sensitive.
 func (p *PrometheusSink) collectAtTime(c chan<- prometheus.Metric, t time.Time) {
+	p.lastCollection.Store(t.UnixNano())
 	expire := p.expiration != 0
 	p.gauges.Range(func(k, v interface{}) bool {
 		if v == nil {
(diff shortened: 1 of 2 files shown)


prometheus/prometheus.go
```diff
@@ -205,6 +210,82 @@ func (p *PrometheusSink) collectAtTime(c chan<- prometheus.Metric, t time.Time)
 	})
 }
 
+// RunBackgroundCleanup starts a background goroutine that periodically removes
+// expired metrics if it's been more than twice the expiration interval since
+// last collection. This ensures metrics are cleaned up even when the endpoint
```
schmichael, line 216: This implies the CPU usage goes up as the expiration interval goes down. I think that's fine, but we'll need to document it in product documentation like we did the memory leak. Not sure there's anything else we can do/document in the library itself.

prometheus/prometheus.go
```diff
@@ -205,6 +210,82 @@ func (p *PrometheusSink) collectAtTime(c chan<- prometheus.Metric, t time.Time)
...
+					p.cleanupExpiredAt(time.Now())
+				}
+			case <-ctx.Done():
+				return
+			}
+		}
+	}()
+}
+
+func (p *PrometheusSink) cleanupExpiredAt(t time.Time) {
```
schmichael: A shame we can't share the loop code between this and `collectAtTime`, but I don't see a way. Maybe add a "if you change this, make sure to check that" style comment here for cases like this where similar logic is expected to be maintained in multiple places?
```suggestion
// cleanupExpiredAt deletes expired metrics. Since collection 
// also deletes expired metrics, changes to one method often
// require changes to the other.
func (p *PrometheusSink) cleanupExpiredAt(t time.Time) {
```

schmichael (commented):
Description nit pick:
> The Prometheus sink clears expired metrics when they're collected. But this means that if a metrics collector ~~fails,~~ _is not running,_ metrics will accumulate in the sink and consume potentially large amounts of memory.

I know it's pedantic, but the recent memory leak I observed occurred because the Prometheus sink was enabled in one deployment with the plan to enable the collector in a subsequent deployment. This seems like an incredibly reasonable approach to operations yet can OOM Nomad agents even without anything else failing.

LGTM but I like your idea of further testing before merging.

prometheus/prometheus.go
tgross: I'm marking the `lastCollection` time on collection calls, so that we can avoid that CPU cost if we're collecting regularly. If we're not collecting regularly, we're only spending the same CPU time that we would be during a regular collection (actually less CPU and memory, because we're not serializing the metrics too).

So yeah I think we still end up needing to document that you won't want to enable Prometheus metrics in downstream products unless you intend to collect, but the cost of that is much much less now.

prometheus/prometheus.go
tgross: I pulled this change in, but then I took another crack at reusing the loop by passing a function to the loop. What do you think of this approach?

prometheus/prometheus.go
schmichael: Of course! Basically [constant work](https://aws.amazon.com/builders-library/reliability-and-constant-work/). Perfect.

Agreed it should still be optional but a much safer option!

prometheus/prometheus.go
schmichael: Hm yeah I think the LoC savings are worth the extra bit of indirection. Add a comment as to what `fn` is and call it a day? (Or not, I don't feel strongly here. Thanks for expirementing.)

schmichael (dismissed):

tgross:
Oops, I committed the linter config that I had staged for a different PR. Will fix.

tgross:
Ok, to test this, I deployed a Nomad cluster with a single server and a single client, comparing the most recent production release (2.0.3) and a build with `go-metrics` replaced by this PR's branch. The Nomad agents were configued with the following telemetry block:

```hcl
telemetry {
  prometheus_metrics                     = true
  publish_node_metrics                   = true
  publish_allocation_metrics             = true
  disable_dispatched_job_summary_metrics = false
  disable_allocation_hook_metrics        = false
}
```

For both versions, I ran a dispatch batch workload of 5000 dispatches staggered over a window of several minutes, ensuring that each child job ran long enough that the client would collect metrics. I took a heap profile of both the server and client at the midpoint of the run, at the end of the run, and 2 minutes after the run was over.

I ran two versions of the test. In the first version, the test runs `nomad system gc` after each step. This clears out dead jobs and allocations so that the only metrics data we're seeing it that retained afterwards. In the second version, the test does not GC. Because the leader emits metrics periodically for job summaries, we expect that we'll see a large chunk of memory devoted to metrics even in the patched agent.

The table below shows the rough results at the end of the run, showing just the heap used for metrics and its percentage of the total.

| Agent          | Heap (no GC) | Heap (with GC) |
|----------------|--------------|----------------|
| 2.0.3 leader   | 186 MB (57%) | 84 MB (46%)    |
| 2.0.3 client   | 170 MB (61%) | 81 MB (11%)    |
| patched leader | 186 MB (57%) | neglible       |
| patched client | neglible     | neglible       |


**2.0.3 leader**

2 0 3-server

**2.0.3 client**

2 0 3-client

**patched leader**

patched-leader

**patched client**

patched-client

schmichael (approved):

PR #207, opened 2026-06-23
CI: update checks and copywrite config

tgross:
This changeset has CI configuration changes to bring it up to speed with the other libraries the Nomad team is maintaining. Includes:
* Update our golangci configuration and fix the resulting issues.
* Remove the long-unused CircleCI and TravisCI configurations.
* Add `copywrite` configuration and update headers that are behind.
* Add a Makefile with the various targets we want for development and CI, and use them in the check workflow.

Ref: https://hashicorp.atlassian.net/browse/NMD-1564


jrasell (approved):
LGTM!

PR #218, opened 2026-08-13
Stabilise TestDisplayMetrics around interval boundaries

### LINKED ISSUE #127: TestDisplayMetrics flaky
Hi!

During automated builds of go-metrics 0.3.4 by the Debian CI infrastructure, we noticed a **test failure** on one of our [armhf](https://wiki.debian.org/ArmHardFloatPort) builder (arm high-float). The issue disappeared after a retry.

All the logs are at https://ci.debian.net/packages/g/golang-github-armon-go-metrics/testing/armhf/, and the two logs of interest here are:
- failure at 2021-09-11 21:15:02 UTC: https://ci.debian.net/data/autopkgtest/testing/armhf/g/golang-github-armon-go-metrics/15197663/log.gz
- success at 2021-09-12 14:55:43 UTC: https://ci.debian.net/data/autopkgtest/testing/armhf/g/golang-github-armon-go-metrics/15198902/log.gz

The lines of interest in the logs:

```
=== RUN   TestDisplayMetrics
    inmem_endpoint_test.go:29: bad: [0xd28240 0xd28400]
--- FAIL: TestDisplayMetrics (0.00s)
```

Looking deeper in the various logs, I noticed that this issue appeared also on the x86_64 architecture:
- 2021-07-25 08:26:55 UTC: https://ci.debian.net/data/autopkgtest/testing/amd64/g/golang-github-armon-go-metrics/13923270/log.gz

Hashim1999164:
Wait for the next in memory interval boundary before recording samples so they land in a single interval.

This matches the flake described in the issue.

Fixes #127

### DIFF
--- a/inmem_endpoint_test.go
+++ b/inmem_endpoint_test.go
@@ -15,8 +15,11 @@ import (
 )
 
 func TestDisplayMetrics(t *testing.T) {
-	interval := 10 * time.Millisecond
+	interval := 20 * time.Millisecond
 	inm := NewInmemSink(interval, 50*time.Millisecond)
+
+	// Wait for the next interval boundary so all samples land in one interval.
+	time.Sleep(time.Until(time.Now().Truncate(interval).Add(interval)))
 
 	// Add data points
 	inm.SetGauge([]string{"foo", "bar"}, 42)
@@ -69,7 +72,7 @@ func TestDisplayMetrics(t *testing.T) {
 					Max:   22,
 					Sum:   42,
 					SumSq: 884,
-					Rate:  4200,
+					Rate:  2100,
 				},
 				Mean:          21,
 				Stddev:        1.4142135623730951,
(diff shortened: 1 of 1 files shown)


tgross (changes requested):
I'm not sure if this is still flaking? Are you having ongoing problems with that in your own CI of this library? If you're just doing drive-by LLM PRs (which seems likely and I'm going to recommend you not), then have you demonstrated this test is still flaky?

In any case, doesn't this just make the flake narrower? Isn't this a case ripe for use of `synctest`?

Hashim1999164:
Sorry for the delay. Taking the questions in order.

This was not failing in a private CI of this library. TestDisplayMetrics buckets with time.Now().Truncate(interval). If a sample lands just before that boundary and the later samples land after it, DisplayMetrics can miss the expected counters. That is a real flake, just a small window, so it rarely shows up unless the test runs a lot.

The first sleep only shrank that window. It did not remove the wall clock. The current commit uses testing/synctest so the fake clock does not advance across a bucket boundary unless the test says so. Every sample stays in one interval, which is the actual fix.

I can still close this if you would rather not take a drive by test change.

